PR #5, opened 2025-09-09
fix(uninstall): fix issues with uninstalling manifest-only drivers

amoeba:
This primarily fixes a bug with dbc uninstall for manifest-only drivers where the extra folder we install alongside the driver manifest wasn't being removed. I also factored out the common logic related to removing the driver shared library and added some defensive checks to limit the impact of a maliciously-crafted driver manifest.

Ref #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,15 @@ curl -LsSf https://dbc.columnar.tech/install.sh | less
 
 #### Windows
 
-Use `irm` to download the script and execute it with `iex`:
+Download the Windows graphical installer for your architecture:
+
+| Architecture |  Installer                                              |
+| ------------ | ------------------------------------------------------- |
+| x64 (64-bit) | <https://dbc.columnar.tech/latest/dbc-latest-x64.msi>   |
+| x86 (32-bit) | <https://dbc.columnar.tech/latest/dbc-latest-x86.msi>   |
+| ARM64        | <https://dbc.columnar.tech/latest/dbc-latest-arm64.msi> |
+
+Or use `irm` to download the install script and execute it with `iex`:
 
 ```sh
 powershell -ExecutionPolicy ByPass -c "irm https://dbc.columnar.tech/install.ps1 | iex


config/config.go
```diff
@@ -315,3 +316,70 @@ func decodeManifest(r io.Reader, driverName string, requireShared bool) (Manifes
...
+			continue
+		}
+
+		// Skip any shared paths with more than one contiguous `.` in them to avoid
+		// path traversal in the middle of a path
+		matched, err := regexp.MatchString("\\.{2,}", sharedPath)
+		if matched || err != nil { // Also ignore errors which shouldn't happen with
+			// such a simple regex
+			continue
+		}
```
zeroshade: should we just use https://pkg.go.dev/path/filepath#Clean?

amoeba: Hadn't seen that, it looks nice! Seems like we should.

config/config.go
```diff
@@ -315,3 +316,70 @@ func decodeManifest(r io.Reader, driverName string, requireShared bool) (Manifes
...
+
+	// Manifest only drivers can come with extra files such as a LICENSE and we
+	// create a folder next to the driver manifest to store them, same as we'd
+	// store the actual driver shared library. Above, we find the path of this
+	// folder by looking at the Driver.shared path. For manifest-only drivers,
+	// Driver.shared is not a valid path (it's just a name), so this trick doesn't
+	// work. We do want to clean this folder up so here we guess what it is and
+	// try to remove it e.g., "somedriver_macos_arm64_v1.2.3."
+	extra_folder := fmt.Sprintf("%s_%s_v%s", info.ID, platformTuple, info.Version)
+	// Return immediately if folder has repeated . chars (for safety)
```
zeroshade: that would only be an issue if the entire text of the folder is `..`, this shouldn't be a problem otherwise, should it?

i.e. `foo_bar..baz` is a perfectly valid filename that shouldn't cause us any problems

amoeba: yeah, this is overly defensive. I can re-work this to use Clean and then catch their special return value (`"."`).

amoeba:
Ready for a review. This now uses Clean on the path and I was able to test the new logic in config.UninstallDriverShared.

zeroshade:
Looks like there's still issues with the windows tests

amoeba:
I'm seeing if I can figure out what's going on in CI by surfacing the error (which I should have done originally).

amoeba:
Hrm, 

> Failed to rename sidecar folder. Something is wrong with this test: rename C:\Users\RUNNER~1\AppData\Local\Temp\TestSubcommandsTestUninstallManifestOnlyDriver3321879900\001\test-driver-manifest-only C:\Users\RUNNER~1\AppData\Local\Temp\TestSubcommandsTestUninstallManifestOnlyDriver3321879900\001\test-driver-manifest-only_windows_amd64_v1.0.0: The system cannot find the file specified.

I may need to log into my Windows machine to see if installing a manifest-only driver works on Windows.

amoeba:
Looks good now @zeroshade.

config/config.go
```diff
@@ -315,3 +314,60 @@ func decodeManifest(r io.Reader, driverName string, requireShared bool) (Manifes
...
+	}
+
+	// Manifest only drivers can come with extra files such as a LICENSE and we
+	// create a folder next to the driver manifest to store them, same as we'd
+	// store the actual driver shared library. Above, we find the path of this
+	// folder by looking at the Driver.shared path. For manifest-only drivers,
+	// Driver.shared is not a valid path (it's just a name), so this trick doesn't
+	// work. We do want to clean this folder up so here we guess what it is and
+	// try to remove it e.g., "somedriver_macos_arm64_v1.2.3."
+	extra_folder := fmt.Sprintf("%s_%s_v%s", info.ID, platformTuple, info.Version)
```
zeroshade, line 363:
```suggestion
	extra_folder := fmt.Sprintf("%s_%s-%s", info.ID, platformTuple, info.Version)
```

cmd/dbc/uninstall_test.go
```diff
@@ -124,3 +125,31 @@ func (suite *SubcommandTestSuite) TestUninstallMultipleLocationsNonDefault() {
...
+		GetModelCustom(baseModel{getDriverList: getTestDriverList, downloadPkg: downloadTestPkg})
+	suite.validateOutput("\r[✓] searching\r\n[✓] downloading\r\n[✓] installing\r\n[✓] verifying signature\r\n"+
+		"\r\nInstalled test-driver-manifest-only 1.0.0 to "+suite.tempdir+"\r\n"+
+		"\r\nMust have libtest_driver installed to load this driver\r\n", suite.runCmd(m))
+	if runtime.GOOS != "windows" {
+		suite.FileExists(filepath.Join(suite.tempdir, "test-driver-manifest-only.toml"))
+	}
+
+	// Verify the sidecar folder exists before we uninstall
+	new_sidecar_path := fmt.Sprintf("test-driver-manifest-only_%s_v1.0.0", config.PlatformTuple())
```
zeroshade, line 140:
```suggestion
	new_sidecar_path := fmt.Sprintf("test-driver-manifest-only_%s-1.0.0", config.PlatformTuple())
```

amoeba:
Thanks for the review @zeroshade!

PR #35, opened 2025-09-15
feat(testing): add cdn test setup with pixi

### LINKED ISSUE #34: Create a tool to create a fake driver index for testing who dbc scales
We currently test dbc against the production driver index and can manually override it to test against the staging driver index. We also can override individual components of dbc using `baseModel`. This works but it would be good to be able to test more flexibly. 

For example, how do we know dbc is fast and will scale well as we add more packages? dbc should be able to support a driver index with 100 packages, each with 100 versions. Probably more.

I think we could create a really basic script that creates a test driver index with fake packages. If we wanted to write tests that tests against this, we could make it deterministic but we could also use the script to do some basic fuzz testing.

amoeba:
Creates a way of manually testing dbc against a CDN full of fake drivers. The number of drivers and number of versions per driver is configurable. This uses pixi just for convenience.

Closes #34.


zeroshade:
Should we just use the manifest and test packages in `cmd/dbc/testdata` as our fake index?

amoeba:
> Should we just use the manifest and test packages in `cmd/dbc/testdata` as our fake index?

We could but I'm not sure we'd want to check in a test scenario like 100 drivers with 100 versions so having a way to dynamically test things is good IMO. I've already found this setup useful for catching issues that aren't covered by the current test suite.

amoeba:
Here's my worst-case scenario test,

```console
$ pixi run cdn-generate -d 100 -v 100
$ wc -l cdn-dev/manifest.yaml
  140701 cdn-dev/manifest.yaml
$ go run ./cmd/dbc search
#  4m 51s goes by...
```

amoeba:
and since dbc install searches first, I see this for at least 4min 51s,

```console
$ go run ./cmd/dbc install yard
[⠋] searching...
[ ] downloading
[ ] installing
[ ] verifying signature```

zeroshade:
Looking into while this is running and hanging, it looks like it actually is the yaml decoding

image

which is ludicrous. It's only 5.4MB of yaml, I'll see if i can figure out what's going on

amoeba:
Thanks for looking! It could be worth the time now to consider switching the serialization format or switching to another YAML package (which I see you're doing already in this PR). Maybe even to a streamable format?

amoeba:
Awesome, the new package is snappy! Merging.

PR #39, opened 2025-09-16
docs: add debugging section to CONTRIBUTING.md

amoeba:
It took me some research to find the specific way to run dlv with dbc so this patch documents it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,9 +28,10 @@ go test -v ./...
 
 ## Debugging
 
-To use the [Delve](https://github.com/go-delve/delve) debugger to debug dbc, some special steps are required.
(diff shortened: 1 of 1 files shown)


amoeba:
> This is great! Maybe we should also mention that much of the work to handle the dlv server is managed for you automatically if you're using VS Code and the Go extension there

Just for tests or even for debugging `go run ./cmd/dbc`? Debugging tests does work out of the box but I was never able to figure out how to debug bubbletea with VS Code so I ended up with this.

zeroshade:
if you use a `launch.json` file you can have it use the debugger directly with a package and args that is the equivalent of `go run ./cmd/dbc`. Don't think we need to add such an update in this PR though

amoeba:
I added a quick first sentence to try to explain that debugging dbc can be done the usual way.

PR #41, opened 2025-09-16
fix: gracefully handle install without signature

zeroshade:
If the `MANIFEST` is missing the `Files.signature` then we first try looking for `driver + '.sig'` and then error out.

Adds a new CLI argument `--allow-missing-signature` which will allow installing if the signature is missing.

### DIFF
--- a/drivers.go
+++ b/drivers.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"runtime/debug"
 	"slices"
@@ -21,6 +22,8 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/ProtonMail/gopenpgp/v3/crypto"
 	"github.com/goccy/go-yaml"
+	"github.com/google/uuid"
+	machineid "github.com/zeroshade/machine-id"
 )
 
 const defaultURL = "https://dbc-cdn.columnar.tech"
@@ -29,6 +32,8 @@ var (
 	baseURL   = defaultURL
 	Version   = "unknown"
 	userAgent string
+	mid       string
+	uid       uuid.UUID
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


cmd/dbc/install.go
```diff
@@ -18,15 +20,17 @@ import (
 
 type InstallCmd struct {
 	// URI    url.URL `arg:"-u" placeholder:"URL" help:"Base URL for fetching drivers"`
-	Driver  string             `arg:"positional,required" help:"Driver to install"`
-	Version *semver.Version    `arg:"-v" help:"Version to install"`
-	Level   config.ConfigLevel `arg:"-l" help:"Config level to install to (user, system)"`
+	Driver       string             `arg:"positional,required" help:"Driver to install"`
+	Version      *semver.Version    `arg:"-v" help:"Version to install"`
+	Level        config.ConfigLevel `arg:"-l" help:"Config level to install to (user, system)"`
+	AllowMissing bool               `arg:"--allow-missing-signature" help:"Allow installation of drivers without a signature file"`
```
amoeba: What about `--no-verify` instead? The `--no-foo` negative form feels a bit more common in some CLI tools.

cmd/dbc/install_test.go
```diff
@@ -126,3 +126,15 @@ func (suite *SubcommandTestSuite) TestInstallManifestOnlyDriver() {
 		suite.FileExists(filepath.Join(suite.tempdir, "test-driver-manifest-only.toml"))
 	}
 }
+
+func (suite *SubcommandTestSuite) TestInstallDriverNoSignature() {
+	m := InstallCmd{Driver: "test-driver-no-sig"}.
+		GetModelCustom(baseModel{getDriverList: getTestDriverList, downloadPkg: downloadTestPkg})
+	out := suite.runCmdErr(m)
+	suite.Contains(out, "signature file 'test-driver-1-not-valid.so.sig' for driver is missing")
```
amoeba, line 135: Should we test the state of the tmpdir after failure so we know what files remain when install fails? It'd be good as a regression test if we change the behavior.

cmd/dbc/install.go
zeroshade: sure

cmd/dbc/install_test.go
zeroshade: Should I make the change in this PR to clean up after the failed install?

PR #43, opened 2025-09-17
feat: rename CDN index file to `index.yaml` (from `manifest.yaml`)

### LINKED ISSUE #21: Rename driver index file to `index.yaml`
We're changing the name of the driver index file on the CDN from `manifest.yaml` to `index.yaml` to avoid confusing it with the driver manifest files. Change it in the dbc code.

lidavidm:
Closes #21.


amoeba:
Thanks! I added a couple more renames that were internal-only. @zeroshade I'll let you merge once you've taken a look.

PR #44, opened 2025-09-17
feat: add a unique id to dbc requests

### LINKED ISSUE #24: Create GUID on first run, send as header with requests to CDN
Add support to dbc for creating a GUID on first run that can be sent with requests to the CDN. We'll need to set up a mechanism for dbc to have its own config dir (separate from ADBC's config dir), store the GUID there, a way to disable sending it, and a page in the docs about it.

zeroshade:
closes #24 

I created github.com/columnar-tech/machine-id to get something a bit better than creating and storing a GUID.

### DIFF
--- a/drivers.go
+++ b/drivers.go
@@ -67,7 +67,12 @@ func init() {
 	}
 
 	// try to read the existing UUID file
-	fp := filepath.Join(userdir, "columnar", "dbc", "uid.uuid")
+	dirname := "columnar"
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		dirname = "Columnar"
+	}
+
+	fp := filepath.Join(userdir, dirname, "dbc", "uid.uuid")
 	data, err := os.ReadFile(fp)
 	if err == nil {
 		if err = uid.UnmarshalBinary(data); err == nil {


drivers.go
```diff
@@ -50,6 +55,36 @@ func init() {
...
+			return
+		}
+	}
+
+	// if the file didn't exist or we couldn't parse it, generate a new uuid
+	// and then write a new file
+	uid = uuid.New()
+	// if we fail to create the dir or write the file, just ignore the error
+	// and use the fresh UUID
+	if err = os.MkdirAll(filepath.Dir(fp), 0o600); err == nil {
```
amoeba:
```suggestion
	if err = os.MkdirAll(filepath.Dir(fp), 0o755); err == nil {
```

we shouldn't create a directory the user can't use

amoeba: with `0o600`, we can never create uid.uuid.

drivers.go
```diff
@@ -50,6 +55,36 @@ func init() {
...
 		}
 	}
+
+	mid, _ = machineid.ProtectedID()
+
+	// get user config dir
+	userdir, err := os.UserConfigDir()
+	if err != nil {
+		// if we can't get the dir for some reason, just generate a new UUID
+		uid = uuid.New()
```
amoeba, line 65: Should we use uuid.Random instead of New? The docs say New panics where as Random does not?

drivers.go
zeroshade: Ah, i forgot you actually do need execute permissions on the folder to read/write the file. Though, should we use `0o700` so that only the user can read/write the folders?

zeroshade: Looks like the default for most things which create directories in my `~/.config` dir use `0o700`

drivers.go
zeroshade: If for some reason we can't create the random UUID, i feel like panic'ing might be the right option as it would indicate that many other things won't work (i.e. anything that relies on randomness such as any cryptographic stuff)

drivers.go
```diff
@@ -50,6 +55,36 @@ func init() {
...
+		return
+	}
+
+	// try to read the existing UUID file
+	fp := filepath.Join(userdir, "dbc", "uid.uuid")
```
ianmcook: Should we nest this under a `Columnar` directory?

drivers.go
ianmcook: FWIW, on my Mac, in `~/Library/Application Support`, 75% of the directories use `0755`

ianmcook: But I can't think of any reason why this would need to be readable by other users, so `0700` seems fine to me.

drivers.go
zeroshade: sure we can

drivers.go
```diff
@@ -50,6 +55,36 @@ func init() {
...
+	// get user config dir
+	userdir, err := os.UserConfigDir()
+	if err != nil {
+		// if we can't get the dir for some reason, just generate a new UUID
+		uid = uuid.New()
+		return
+	}
+
+	// try to read the existing UUID file
+	fp := filepath.Join(userdir, "columnar", "dbc", "uid.uuid")
```
ianmcook: On Windows and macOS, the convention is to use capital letters for vendor name directories here (`Columnar`). On Linux the convention is lowercase (`columnar`). Can we follow that convention here?

`dbc` should be lowercase everywhere.

drivers.go
amoeba: I guess that sounds fine.

drivers.go
amoeba: 0o700 seems like it should be enough. On my Linux machine ~/.config is all 777. 🤷

PR #46, opened 2025-09-17
fix: wrap post-install message text

### LINKED ISSUE #19: Fix issue with truncated post-install message
dbc install doesn't print the entire post install message here,

```toml
...
[PostInstall]
messages = ["To use this driver, you must download the libduckdb library for your platform from https://github.com/duckdb/duckdb/releases/tag/v1.3.2 and make it available in your .so/.dylib/.dll import path."]
```

I only get

> To use this driver, you must download the libduckdb library for your platform from https://github.com/duckdb/duckdb/releases/tag

zeroshade:
closes #19

### DIFF
--- a/cmd/dbc/uninstall_test.go
+++ b/cmd/dbc/uninstall_test.go
@@ -129,9 +129,9 @@ func (suite *SubcommandTestSuite) TestUninstallMultipleLocationsNonDefault() {
 func (suite *SubcommandTestSuite) TestUninstallManifestOnlyDriver() {
 	m := InstallCmd{Driver: "test-driver-manifest-only"}.
 		GetModelCustom(baseModel{getDriverList: getTestDriverList, downloadPkg: downloadTestPkg})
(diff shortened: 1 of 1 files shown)


amoeba:
I bet this fixes the direct issue but I think there's a fundamental issue we might want to address since this happens elsewhere if you look closely. See how this cuts off?

Screenshot 2025-09-17 at 2 19 01 PM

zeroshade:
Hmm, good point. Let me see if there's any general option I can pass to bubbletea to handle this

amoeba:
I noticed that the charmbracelet team has an example where we add a "finalOutput" member to the cmd struct (alongside status) and then have the root cmd print the finalOutput. I started on this in https://github.com/amoeba/dbc/commit/58d080393d409756c3ed278415b6d12554bb101e.

amoeba:
See https://github.com/charmbracelet/bubbletea/blob/main/examples/result/main.go.

zeroshade:
Ah, you mean shift the output for when we're *done* to **after** the return instead of as part of the `View` method?

amoeba:
Yep. It cleanly separates the output into two categories. Requires quite a bit of change I think.

Issue #29, opened 2025-09-13
uninstall erroring and deleting more than it should


amoeba:
On macOS, dbc v0.0.1-pre3 is erroring unexpectedly when I uninstall a driver and it also looks like it deletes more than it should. This bug may be fixed or at least affected by https://github.com/columnar-tech/dbc/pull/5 but I haven't tested yet.

```sh
$ ls "/Users/bryce/Library/Application Support/ADBC/Drivers"
ls: /Users/bryce/Library/Application Support/ADBC/Drivers: No such file or directory

~/src/columnar-tech/dbc docs/update-website*
$ dbc install mssql
[✓] searching
[✓] downloading
[✓] installing
[✓] verifying signature

Installed mssql 1.0.0 to /Users/bryce/Library/Application Support/ADBC/Drivers

~/src/columnar-tech/dbc docs/update-website*
$ dbc search
• duckdb - An analytical in-process SQL database management system
• snowflake - An ADBC driver for Snowflake developed under the Apache Software Foundation
• mssql - Columnar ADBC Driver for Microsoft SQL Server [installed: user=>1.0.0]
• mysql - ADBC Driver Foundry Driver for MySQL
• flightsql - An ADBC driver for Apache Arrow Flight SQL developed under the Apache Software Foundation

~/src/columnar-tech/dbc docs/update-website*
$ ls "/Users/bryce/Library/Application Support/ADBC/Drivers"
mssql_macos_arm64_v1.0.0/ mssql.toml

~/src/columnar-tech/dbc docs/update-website*
$ dbc uninstall mssql
Error: failed to uninstall driver: error removing manifest /Users/bryce/Library/Application Support/ADBC/Drivers/mssql.toml: remove /Users/bryce/Library/Application Support/ADBC/Drivers/mssql.toml: no such file or directory

~/src/columnar-tech/dbc docs/update-website*
$ ls "/Users/bryce/Library/Application Support/ADBC/Drivers"
ls: /Users/bryce/Library/Application Support/ADBC/Drivers: No such file or directory

~/src/columnar-tech/dbc docs/update-website*
```

amoeba:
This was caused by what was going on in https://github.com/columnar-tech/dbc/issues/37 and so this no longer happens with the updated mysql package.

I think we still need to make the uninstall routine more defensive. I'll make a test for this.

PR #48, opened 2025-09-17
fix: add test to cover behavior for invalid manifests

amoeba:
Fixes an edge case where we get confused about what fields to put in a manifest. When we package a driver, we include an all-caps `MANIFEST` file which has a `Files` table that points to the shared library living in the same package. When dbc installs a package, it converts what's in `Files` into what goes in `Driver.shared`. If we goof up and put a `Driver.shared` table in `MANIFEST`, `dbc uninstall` was mistakenly deleting _all_ drivers in the config level.

This patch catches this case and handles it. It's good to be defensive here because we'll be asking people to crate `MANIFEST` files who aren't us so mistakes will happnen.

Closes https://github.com/columnar-tech/dbc/issues/29


cmd/dbc/testdata/test_manifest.yaml
```diff
@@ -60,3 +60,16 @@ drivers:
...
           - platform: windows_amd64
+  - name: Test Driver Invalid Manifest
+    description: This is test driver with an invalid manfiest. See https://github.com/columnar-tech/dbc/issues/37.
+    license: Apache-2.0
+    path: test-driver-invalid-manifest
+    pkginfo:
+      - version: v1.0.0
+        packages:
+          - platform: linux_amd64
+          - platform: linux_ard64
```
zeroshade:
```suggestion
          - platform: linux_arm64
```

amoeba:
Okay good. I rebased this and now it's failing like I expect. The current code is exposing a bad bug we need to patch. If Driver.shared isn't a file like it's supposed to be, we actually delete the entirety of `ADBC_DRIVER_PATH`, meaning we nuke all of the user's drivers.

amoeba:
I ended up just explicitly detecting the weird state that causes this bug, see https://github.com/columnar-tech/dbc/pull/48/commits/43bd381c69b96a2c48d16f441e83edaf6108b256. It might be cleaner just entirely fail the install when we get a MANIFEST that has this issue but until then this catches it.

cmd/dbc/uninstall_test.go
```diff
@@ -153,3 +154,50 @@ func (suite *SubcommandTestSuite) TestUninstallManifestOnlyDriver() {
...
+	// Assert that it's a folder
+	suite.DirExists(value)
+	// and continue
+
+	m = UninstallCmd{Driver: "test-driver-invalid-manifest", Level: config.ConfigEnv}.GetModel()
+	output := suite.runCmd(m)
+
+	// TODO: Currently failing with:
+	// Error: failed to uninstall driver: error removing driver test-driver-invalid-manifest: RemoveAll .: invalid argument
+	suite.validateOutput("Driver `test-driver-invalid-manifest` uninstalled successfully!\r\n\r\n\r ", "", output)
```
zeroshade: the tests currently pass, so this comment isn't correct?

config/config.go
```diff
@@ -371,3 +400,34 @@ func UninstallDriverShared(cfg Config, info DriverInfo) error {
 
 	return nil
 }
+
+// Determine whether target is an immediate subdirectory of base.
+func IsImmediateSubDir(base string, target string) (bool, error) {
```
zeroshade: can we avoid exporting this?

config/config.go
```diff
@@ -331,7 +331,36 @@ func UninstallDriverShared(cfg Config, info DriverInfo) error {
...
+			//     └── libadbc_driver_linux_amd64_v1.0.0.so
+			//
+			// If the MANIFEST we're installed has Driver.shared already set, this is
+			// invalid becuase it should have Files set instead. When this happens,
+			// the Driver.shared value in the installed manifest ends up as a absolute
+			// path to a folder, not a shared library.
+			shared_path_is_subdir, err := IsImmediateSubDir(cfg.Location, sharedPath)
+			if err != nil {
+				return fmt.Errorf("error removing driver %s: %w", info.ID, err)
+			}
```
zeroshade: Could we simplify this logic by using `filepath.Ext(sharedPath)` and assuming that it should have a non-empty extension? (`so`, `dll`, `dylib`, etc.)

We could also leverage https://pkg.go.dev/os#Root.FS to ensure that we don't have any issues with malicious symbolic links. Maybe do `os.OpenRoot(cfg.Location)` and then work in terms of the returned root FS?

cmd/dbc/uninstall_test.go
amoeba: My mistake here, removed in 1b43728.

config/config.go
amoeba: Yeah I should have done that. Fixed in 49cf0fd.

config/config.go
amoeba: > Could we simplify this logic by using filepath.Ext(sharedPath) and assuming that it should have a non-empty extension? (so, dll, dylib, etc.)

I don't know if that would quite cover it. We basically just want to know whether calling `filepath.Dir` doesn't give us a subfolder of cfg.Location. It should almost every time, just not in this edge case.

https://pkg.go.dev/os#Root.FS looks cool. That could be a nice improvement (followup?).

zeroshade: > I don't know if that would quite cover it. We basically just want to know whether calling filepath.Dir doesn't give us a subfolder of cfg.Location. It should almost every time, just not in this edge case.

Should we just do `info, _ := os.Stat()` and `info.IsDir()` then? would be simpler

amoeba: I'm not sure I see how that'd catch the issue. If you want to take a shot at the idea I can take a look. Knowing the path is a folder is part of detecting that we're in the edge case but we also need to ensure it's a folder that, when we call `filepath.Dir`, we don't get `cfg.Location` and then delete all the user's drivers.

zeroshade: Could we just do `filepath.Clean(dir) != filepath.Clean(cfg.Location)` then?

amoeba: Yeah I guess so. Maybe we don't need to also remove the folder in the case of the edge case.

zeroshade: i like that better in the edge case, so we can reduce the complexity of the logic

amoeba: Changed.

config/config.go
```diff
@@ -331,7 +331,16 @@ func UninstallDriverShared(cfg Config, info DriverInfo) error {
...
 		if info.Source == "dbc" {
-			if err := os.RemoveAll(filepath.Dir(sharedPath)); err != nil {
+			sharedDir := filepath.Dir(sharedPath)
+			// Edge case when manifest is ill-formed: if sharedPath is set to the
+			// folder containing the shared library instead of the shared library
+			// itself, sharedDir is cfg.Location and we definitely don't want to
+			// remove that
+			if sharedDir == cfg.Location {
+				continue
+			}
```
zeroshade, line 341: Looks good to me, I think doing a followup where we leverage `os.OpenRoot` will cover the only other edge case I'm worried about (a malicious manifest that points to an entirely different directory)

amoeba: Sounds good. I filed a follow-on https://github.com/columnar-tech/dbc/issues/55.

PR #60, opened 2025-09-20
fix: add test for environment variable precedence

amoeba:
I saw while writing docs that we don't have an explicit test of the precedence chain we document so I created one. However, it current fails in a very strange way for me so this PR adds a failing test.


amoeba:
This test fails early on and when I poke around the `suite.tempdir`, I see a situation that looks like mangled paths?

```
$ tree
.
├── driver_path
│   └── test-driver-1.1
│       ├── test-driver-1-not-valid.so
│       └── test-driver-1-not-valid.so.sig
└── driver_path:
    └── tmp
        └── TestSubcommandsTestInstallEnvironmentPrecedence1409964535
            └── 001
                └── venv_path
                    └── etc
                        └── adbc
                            └── drivers:
                                └── tmp
                                    └── TestSubcommandsTestInstallEnvironmentPrecedence1409964535
                                        └── 001
                                            └── conda_path
                                                └── etc
                                                    └── adbc
                                                        └── drivers
                                                            └── test-driver-1.toml
```

PR #74, opened 2025-09-23
chore: add pre-commit workflow to ci for license checks

amoeba:
Follow-on to https://github.com/columnar-tech/dbc/pull/73.

Adds a new GHA workflow to run pre-commit and a pre-commit config that runs https://github.com/amoeba/license-header-check/. Includes fixes to files that were missing licenses.

### DIFF
--- a/.github/workflows/dev.yaml
+++ b/.github/workflows/dev.yaml
@@ -34,14 +34,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
-      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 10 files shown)


amoeba:
As a follow-on, we could:

1. Add more pre-commit checks that make sense
2. Gate our main workflows on the pre-commit workflow if others like that

amoeba:
Rebased the pr and updated the workflow. The pre-commit only checks to make sure we spell our name right at the moment. I did this assuming Apache RAT can check that the entire license is right. cc @zeroshade

.github/workflows/dev.yaml
```diff
@@ -0,0 +1,50 @@
...
+  contents: read
+
+jobs:
+  lint:
+    name: "Run pre-commit"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-python@v6
```
zeroshade: use sha commit hashes instead of the tags

amoeba: Good call. Done in https://github.com/columnar-tech/dbc/pull/74/commits/e3c883eeccb1d275723ba3cbf833e1ca011faf76. 

Do you know of a good way to manage this automatically for a repo?

PR #78, opened 2025-09-23
docs: add custom javascript to remove prompt and output

amoeba:
Creates and enables an extra.js script adapted from uv's docs that cleans copied code blocks by removing any prompts or output. Also updates some areas in the docs where we weren't using prompt chars that we now can.

Any code blocks with prompt characters must be `console` not `sh` for this to work.

Closes https://github.com/columnar-tech/dbc/issues/54

### DIFF
--- a/docs/scripts/extra.js
+++ b/docs/scripts/extra.js
@@ -29,7 +29,7 @@ function cleanupClipboardText(targetSelector) {
   //     <span>...</span>
   // </code>
 
-  return Array.from(targetElement.childNodes)   // <-- array of lines
+  const lines = Array.from(targetElement.childNodes)   // <-- array of lines
     .map((span_el) => {
       const segments = Array.from(span_el.childNodes).filter(    // <-- array of segments
         (node) => !excludedClasses.some((excludedClass) =>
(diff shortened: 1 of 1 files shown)


ianmcook:
Very cool! It might be safer to use `&&` instead of `;`.

OTOH I think `;` will generally work on Windows in `cmd.exe` and PowerShell but `&&` won't.

ianmcook:
I clicked the button to copy the YAML at `concepts/driver_manifest` and it replaced the line breaks with semicolons.

ianmcook:
Looks like this doesn't handle `#` comment characters properly (like in the blocks on the `guides/driver_list` page). But the solution might be to get rid of the `#` characters there.

amoeba:
Thanks for testing. I also notice that pycon blocks aren't perfect. Fixing these things now.

amoeba:
Re: `;` vs `&&`, I went for `;` for being as cross plat as possible since, as you note, not all shells support `&&` (even fish didn't for a while). However, I think we can maybe get away with preserving the line breaks. So when a user copies a multi-command code block, they get,

```
first command
second command
third command
```

bash and zsh don't like this (though fish handles it gracefully) but I don't think it's an unreasonable choice on our part: The user sees multiple lines of commands and might be expecting their clipboard gets filled with similarly-structured output (i.e., we should preserve whitespace where we can and try to just remove prompt/comment chars).

amoeba:
I've done some manually testing and I think the only place this doesn't work great is pycon blocks (we don't preserve whitespace which python needs to parse). This could be hard to do without rewriting the code. Can we skip it?

ianmcook:
With an assist from Claude Code, I fixed the leading whitespace issue and also made the script preserve blank lines. @amoeba please take a look.

amoeba:
Seems to work nicely, thanks @ianmcook.

amoeba:
Tested in Safari and Firefox.

PR #92, opened 2025-09-24
fix: add workaround for driver manager venv path bug

amoeba:
This works around a bug we identified with the 1.8.0 driver managers when using virtual environments. They are supposed to locate drivers in `$VIRTUAL_ENV/etc/adbc/drivers` but they look in `$VIRTUAL_ENV/etc/adbc/`. There may be a patch release soon to fix this but, in the mean time, we can work around this by symlinking manifests from `$VIRTUAL_ENV/etc/adbc/drivers` to `$VIRTUAL_ENV/etc/adbc/`. The manifests use absolute paths so it's fine.

This PR also adds an integration workflow which is how I tested my change. I don't think we want to run it on every single commit to the repo so we might discuss here when we might want to run something like this or if we want to run it in another repo.


amoeba:
One thing to think about..

There are two reasonable approaches here: Create a symlink or duplicate the file. The symlink approach seems to work on Linux and I expect it will work on macOS but, on Windows, symlinks a trickier. Windows normally requires admin privs to create symlinks but Go supports detection of "Developer Mode" (an OS-wide setting IIRC) which lets non-admin users create symlinks.

We could fall back to duplicating the manifest instead of symlinking on Windows.

amoeba:
At least in CI (which I bet has developer mode turned on), the symlink approach seems to work fine. I'll keep working on this tomorrow.

amoeba:
I'm happy with this but as a final test, I've commented out the code change and am letting CI test.

amoeba:
Okay this looks good. Without this change, the GHA workflow fails each OS:

Linux fails: `additional search path: /home/runner/work/dbc/dbc/.venv/etc/adbc`
macOS fails: `additional search path: /Users/runner/work/dbc/dbc/.venv/etc/adbc`
Windows fails: `additional search path: D:\a\dbc\dbc\.venv\etc/adbc`

See the associated run: https://github.com/columnar-tech/dbc/actions/runs/18027435968/job/51297285706.

amoeba:
Great, when I re-enable the workaround, all CI jobs pass, see https://github.com/columnar-tech/dbc/actions/runs/18027487417.

amoeba:
Ready for 👀 @zeroshade

.github/workflows/integration.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          python-version: "3.13"
+
+      - name: Build dbc
+        run: |
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            go build -o dbc.exe ./cmd/dbc
+          else
+            go build -o dbc ./cmd/dbc
+          fi
```
zeroshade, line 59: This can be simplified as just `go build ./cmd/dbc` it will automatically use `dbc` or `dbc.exe` without us needing to maintain the logic with `-o`

ci/scripts/run_integration.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -eux
+
+python -m venv .venv
+
+if [ -f ".venv/bin/activate" ]; then
+  . ".venv/bin/activate"
+else
+  . ".venv/Scripts/activate"
+fi
```
zeroshade, line 11: should we rename this better as `run_venv_integration`?

scripts/install.sh
```diff
@@ -1111,6 +1115,7 @@ install() {
         add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" ".zshrc .zshenv" "sh"
         exit3=$?
         # This path may not exist by default
+        echo "FISH"
```
zeroshade, line 1118: should we remove this?

.github/workflows/integration.yml
amoeba: Oh, nice. I had a heck of a time finding the right incantation to run dbc on Windows in CI. I'll change this.

scripts/install.sh
```diff
@@ -1097,10 +1097,14 @@ install() {
...
 
     # Avoid modifying the users PATH if they are managing their PATH manually
+    echo "NO_MODIFY_PATH: $NO_MODIFY_PATH"
+    echo "$PATH"
+    echo "$_install_dir"
     case :$PATH:
       in *:$_install_dir:*) NO_MODIFY_PATH=1 ;;
          *) ;;
     esac
+    echo "NO_MODIFY_PATH: $NO_MODIFY_PATH"
```
zeroshade, line 1107: was it intentional to leave the echos here?

scripts/install.sh
amoeba: oh, yep. Not sure how I ended up with that.

scripts/install.sh
amoeba: this was unintentional, I need to back out all the changes to install.sh

amoeba:
Hrm, not sure what closed the PR.

PR #102, opened 2025-09-26
fix: add workaround for driver manager venv path bug

amoeba:
This works around a bug we identified with the 1.8.0 driver managers when using virtual environments. They are supposed to locate drivers in `$VIRTUAL_ENV/etc/adbc/drivers` but they look in `$VIRTUAL_ENV/etc/adbc/`. There may be a patch release soon to fix this but, in the mean time, we can work around this by symlinking manifests from `$VIRTUAL_ENV/etc/adbc/drivers` to `$VIRTUAL_ENV/etc/adbc/`. The manifests use absolute paths so it's fine.

This PR also adds an integration workflow which is how I tested my change. I don't think we want to run it on every single commit to the repo so we might discuss here when we might want to run something like this or if we want to run it in another repo.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Build dbc
         run: |
-          go build -o dbc ./cmd/dbc
+          go build  ./cmd/dbc
         shell: bash
 
       - name: Run Integration Tests
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Build dbc
         run: |
-          go build  ./cmd/dbc
+          go build ./cmd/dbc
         shell: bash
 
       - name: Run Integration Tests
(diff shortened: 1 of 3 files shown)


amoeba:
Changing the repo visibility broke the link between the main repo and my fork so I had to re-create https://github.com/columnar-tech/dbc/pull/92.

amoeba:
@zeroshade Re: https://github.com/columnar-tech/dbc/pull/92#pullrequestreview-3273633502

> should we rename this better as run_venv_integration?

I think I'd rather keep it for now. Over time if we expand what tests are done I imagine we can change this.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,60 @@
...
+          cache-dependency-path: go.sum
+
+      - name: Install Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+
+      - name: Build dbc
+        run: |
+          go build -o dbc ./cmd/dbc
```
zeroshade: are we using the `-o` so that it doesn't call it `dbc.exe` on windows? otherwise you can just remove the `-o` argument entirely since by default it'll use `dbc` as the name since that's the package name.

amoeba: Ah, nice. I just always run it with -o. Removed in 30d3b4c56a3ea6665f7c42c600450f5596414723 and will watch CI.

amoeba:
Thanks. Once CI passes I'll merge.

Issue #62, opened 2025-09-21
Set up PR deploy previews for docs site


amoeba:
It would be good to have deploy previews in docs PRs for reviewers to look at.

We're currently using GitHub pages to deploy the site from `main`. Prior to this, we had PR deploy previews set up but we broke this when we started publishing from main. As far as I can tell, you can't both:

1. Deploy GitHub pages from main to a domain name
2. Deploy private previews

GitHub has this on their roadmap but I think our only option is to use something like Netlify. We should set that up.

amoeba:
I tried two approaches to doing this today with Netlify and it looks like Netlify won't work for us. There are two approaches available to us:

1. Pull-based: Netlify pulls our repo in and builds automatically
2. Push-based: We use GHA and an API request to trigger a build

For (1), the default behavior is that it will always add a noisy comment in PRs about the deploy, which I don't want. The closest you can get is to set an ignore rule so the build only happens when certain paths have been changed in the PR but the comment is still created.

For (2), Netlify doesn't have an official GitHub Actions workflow nor does it have scoped authentication tokens. So we have to use a third-party workflow and give it access to a PAT (read: account-wide) via a GitHub Actions Secret. This opens us up to risk if we ever do more with our Netlify account since a PR could pwn the token. Secrets are (for good reason) inaccessible to forks but forks are exactly the types of PR sources I wanted to enable PR deploy previews for. It looked like there was a way to use Environments and Secrets in combination but I wasn't able to find a way to make it work with forks. We could do (2) but we'd be restricted to pull requests from the main dbc repo which is just an annoying limitation.

I'll leave this ticket open because there are other approaches we could consider using other services.

amoeba:
I moved this out of 0.1 since I don't think we can or need to solve this yet.

PR #104, opened 2025-09-26
chore: add workflow to preview docs PRs on Netlify

amoeba:
Creates a new workflow that publishes the docs to Netlify for PRs that touch paths related to the docs. Some things to note:

- When the deploy runs, a command gets added to the PR (see below)
- New comments aren't made when the preview is re-deployed
- The workflow also attaches a "deploy" to the PR so you can click the "View deployment" button (see below) to preview the page
- This is currently limited to running only on PRs not made from forks. i.e., if you want to update the docs and get previews, you have to push to columnar-tech, not a fork. This is because of the default behavior of secrets and Actions. If we wanted to relax this, we'd need to set up an Environment (I don't have access).

Closes https://github.com/columnar-tech/dbc/issues/62

### DIFF
--- a/.github/workflows/docs_preview.yml
+++ b/.github/workflows/docs_preview.yml
@@ -2,12 +2,9 @@ name: Deploy Docs Preview
 
 on:
   pull_request:
-    # paths:
-    #   - "src/**"
-    #   - "public/**"
-    #   - "package.json"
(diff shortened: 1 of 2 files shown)


amoeba:
I'm going to try out environments, closing so I can test.

PR #106, opened 2025-09-26
draft: netlify pr deploys

amoeba:


amoeba:
I think we can't test this easily until we've merged. I did a fair bit of looking around and it looks like you just can't access Environment secrets (which we're using) if the workflow is `pull_request`. So I'm trying `pull_request_target` which should be able to access Environment secrets. But we can't test that here because GitHub doesn't run `pull_request_target` workflows on the PRs that add them I guess.

.github/workflows/docs_preview.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull-requests: write
+  deployments: write
+  statuses: write
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: "deploy-to-netlify"
+    steps:
+      - uses: actions/checkout@v4
```
lidavidm, line 20: AIUI, on a pull_request_target trigger this checks out the repository on the target branch and not the pull request (quite intentionally), so this won't do what you want.

lidavidm (commented):
I guess we're protected by needing approval but this is kind of a footgun with `pull_request_target`.

.github/workflows/docs_preview.yml
amoeba: ah, okay. The docs around this use "in the context" which is a thing I don't know what it is and I couldn't tell which checkout you get.

lidavidm: https://github.com/orgs/community/discussions/22363#discussioncomment-3236409 is probably a better summary than the docs themselves (oops)

amoeba: nice, thanks

amoeba:
I'm going to close this for now. The risks and complexity aren't worth the benefits.

PR #107, opened 2025-09-27
draft: add netlify.toml

amoeba:


amoeba:
This didn't work as expected but it could be because Netlify may not sync the netlify.toml I add here until it's merged. I'll do some testing privately and return to this.

amoeba:
Okay so this also doesn't work. You can set up a netlify.toml and make it so it doesn't do a deploy unless certain paths have been changed but Netlify will _always_ comment and the comment will say the build was cancelled. So there's no way to set Netlify up in a pull based fashion.

PR #109, opened 2025-09-27
docs: use uv for docs build setup

### LINKED ISSUE #108: Use Python 3.12 for building docs
`mkdocs serve` doesn't work for me with Python 3.13.

ianmcook:
Closes #108

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,9 +24,9 @@ Building the docs requires a Python installation.
 From the root of the repository, run:
 
 ```sh
-python3 -m venv .venv
+uv venv .venv --python python3.12
 source .venv/bin/activate
-pip install -r docs/requirements.txt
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -24,9 +24,9 @@ Building the docs requires a Python installation.
 From the root of the repository, run:
 
 ```sh
-python3 -m venv .venv
+uv venv .venv --python python3.12
```
amoeba: why 3.12?

ianmcook: `mkdocs serve` doesn't work for me with Python 3.13.

amoeba: hrm, weird. works fine here. I'd rather not have the flag in the docs unless we know something really doesn't work.

ianmcook: weird. removed.

PR #112, opened 2025-09-27
docs: shorten go one-liner

ianmcook:
Using a two-character alias in the import makes the following line just short enough to fit without horizontal scroll on desktop at normal zoom

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -31,9 +31,9 @@ dbc is a command-line tool for installing and managing [ADBC](https://arrow.apac
     === "Go"
 
         ```go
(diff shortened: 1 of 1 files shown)


ianmcook:
@zeroshade using dot import would make this even shorter:

```go
import . "github.com/apache/arrow-adbc/go/adbc/drivermgr"

db, _ := Driver{}.NewDatabase(map[string]string{"driver": "mssql", ... })
```

Do you think that's OK in an example context like this?

zeroshade:
That's fine, or you can just make it two lines....

PR #119, opened 2025-10-09
docs: add missing close quote in Windows install command

ianmcook:


ianmcook:
CI failure is unrelated

PR #124, opened 2025-10-13
ci(release): add msi and packages to blob upload

zeroshade:


zeroshade:
they are already defined (the values refer to the `id` property in the configurations. so `nfpms` has `id: pkgs` etc.

amoeba:
ah, got it. thanks

PR #125, opened 2025-10-13
feat(install): add progress bar

### LINKED ISSUE #99: Add progress bar to install subcommand
We should add a progress bar of some sort to dbc install.

Our current subcommand shows the steps it takes,

```
$ dbc install duckdb
[✓] searching
[✓] downloading
[✓] installing
[✓] verifying signature
```

A progress bar would be a nice improvement since anything over ~300KB will take time on slower connections. duckdb is the best example of a slow-to-download driver.

zeroshade:
closes #99


zeroshade:
@amoeba give it a try now :smile:

PR #127, opened 2025-10-13
docs: add pypi project links

### LINKED ISSUE #115: Add project links to PyPI
Like other packages on PyPI have:
Image

zeroshade:
closes #115 

Also needed to change the logo image in the README.md to be an absolute link so that it shows up correctly on PyPi (currently the image is broken because it's not an absolute link).


README.md
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-# dbc <picture><img src="resources/dbc_logo_animated_padded.png?raw=true" width="180" align="right" alt="dbc Logo"/></picture>
+# dbc <picture><img src="https://raw.githubusercontent.com/columnar-tech/dbc/refs/heads/main/resources/dbc_logo_animated_padded.png?raw=true" width="180" align="right" alt="dbc Logo"/></picture>
```
amoeba, line 17: I notice in uv's readme they use a GitHub "asset" for the image in their PyPI readme:

`https://github.com/astral-sh/uv/assets/1309177/03aa9163-1c79-4a87-a31d-7a9311ed9310. I'm not entirely sure how they created it. See the [blame](https://github.com/astral-sh/uv/commit/01b7a27be42b3673f25e0ab64dfe1a6cebcc3a2f#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

raw.githubusercontent may be fine though, what do you think?

scripts/create_wheels.py
```diff
@@ -288,6 +288,13 @@ def dummy(): """Dummy function for an entrypoint. dbc is executed as a side effe
             ("License-Expression", "Apache-2.0"),
             ("License-File", "LICENSE"),
             ("Requires-Python", "~=3.5"),
+            ("Project-URL", "homepage, https://columnar.tech/dbc"),
```
amoeba, line 291: do we want to make this URL real?

zeroshade: Oh? we didn't? lol 

@ianmcook what URL should be the homepage?

README.md
amoeba: I think all they did here was drag-and-drop a file into a comment and use that URL. I was initially confused because the URL is slightly different but that's just because github changed their URL structure at some point and uv must have done this prior?

I think a raw link is probably fine as long as we think we'll notice if we change the filename. PyPI seems to put images behind their own CDN anyway, for example uv's benchmark chart is actually served from this URL on their PyPI readme:

```
https://pypi-camo.freetls.fastly.net/a68fe262b094faba6c9fab4b2d4dbf135709916c/68747470733a2f2f6769746875622e636f6d2f61737472616c2d73682f75762f6173736574732f313330393137372f36323965353963302d396336652d343031332d396164342d616462326263663530383064.
```

zeroshade: I'm not entirely sure how they created it either.... that would definitely be a good idea if we can figure it out

amoeba: I just did the above thing here and came up with,

https://github.com/user-attachments/assets/09243a94-4c29-47da-8b85-ddb6980ad4cd

Either way (with raw or this method) we're abusing GitHub for asset storage and I can't find any great docs. Generally, it seems like user-attachments stick around for quite a long time.

zeroshade: I'd rather point at the `raw.githubusercontent.com` which specifically points to the file in the repo which I think is safer

amoeba: Sounds good.

scripts/create_wheels.py
ianmcook: This URL will be real shortly. Let's keep it as is.

PR #130, opened 2025-10-16
fix: ignore dbc binary in .gitignore

ianmcook:


ianmcook:
@amoeba are the CI failures unrelated?

amoeba:
Yeah, we have a persistent test flake so any test failures that look like that can be ignored.

PR #134, opened 2025-10-17
fix: support entrypoint on Windows

### LINKED ISSUE #131: dbc not setting entrypoint on Windows
While testing on Windows, I noticed that the Python driver manager wasn't able to load the duckdb driver with just driver="duckdb". When I looked in the registry, I didn't see the `entrypoint` set to the expected "duckdb_adbc_init". The `MANIFEST` in our published driver package looks fine,

```
name = "DuckDB"
publisher = "DuckDB Foundation"
license = "MIT"
version = "v1.4.0"

[ADBC]
version = "v1.1.0"

[Driver]
entrypoint = "duckdb_adbc_init"    <---- this is what I expect

[Files]
driver = "duckdb.dll"
signature = "duckdb.dll.sig"
```

So I think we need to fix this in dbc.

amoeba:
Closes #131


amoeba:
Tested this manually on my Windows machine with the latest duckdb driver from the production driver index and it works.

PR #143, opened 2025-10-20
ci: add homebrew release

### LINKED ISSUE #85: Support Homebrew installation for dbc
Currently, dbc supports installation via PyPI (pip, pipx), standalone installers, system packages (.deb, .rpm, .snap), and Docker images. However, there is no official Homebrew formula or cask for macOS users.

zeroshade:
closes #85

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -305,8 +305,7 @@ snapshot:
 homebrew_casks:
   - name: dbc
     alternative_names:
-      - dbc@{{ .Version }}
-      - dbc@{{ .Major }}
+      - dbc@{{ .Version }}      
       - dbc@{{ .Major }}.{{ .Minor }}
     ids:
       - archives
@@ -321,10 +320,10 @@ homebrew_casks:
     description: "A CLI utility for managing ADBC drivers"
     repository:
       owner: "columnar-tech"
-      name: "homebrew-dbc"
+      name: "homebrew-tap"
       branch: main
       token_type: "github"
-      token: "{{ .ENV.GITHUB_TOKEN }}" # we'll probably need a dedicated PAT or something?
+      token: "{{ .ENV.DBC_TAP_GITHUB_PAT }}" # we'll probably need a dedicated PAT or something?
 
 release:
   ids:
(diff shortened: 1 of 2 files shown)


zeroshade:
I haven't created `columnar-tech/homebrew-dbc` yet. I'll do so once we're clear on this PR

.goreleaser.yaml
```diff
@@ -288,6 +288,30 @@ notarize:
 snapshot:
   version_template: "{{ incpatch .Version }}-snapshot"
 
+homebrew_casks:
+  - name: dbc
+    alternative_names:
+      - dbc@{{ .Version }}
+      - dbc@{{ .Major }}
+      - dbc@{{ .Major }}.{{ .Minor }}
```
amoeba:
```suggestion
      - dbc@{{ .Version }}
      - dbc@{{ .Major }}.{{ .Minor }}
```

I think we could totally skip this but at least I don't think we need all three aliases. What do you think about just Major.Minor for now?

.goreleaser.yaml
```diff
@@ -288,6 +288,30 @@ notarize:
...
+      verified: "github.com/columnar-tech/dbc/"
+      using: ":homebrew_curl"
+      cookies:
+        license: "accept-backup"
+    commit_msg_template: "Brew cask update for {{ .ProjectName }} version {{ .Tag }}"
+    homepage: "https://columnar.tech/dbc"
+    description: "A CLI utility for managing ADBC drivers"
+    repository:
+      owner: "columnar-tech"
+      name: "homebrew-dbc"
```
amoeba:
```suggestion
      name: "homebrew-tap"
```

It's slightly more conventional to call the repo "homebrew-tap". That carries the benefit of us not having to rename the repo if we create more formulae/casks.

zeroshade: Is it? Everything I was seeing was that it should be `homebrew-${package}` I thought? Though I guess you're right for the case where we might want to create more under the Columnar org...

@ianmcook thoughts?

amoeba: I mean: https://github.com/charmbracelet/homebrew-tap. :) but also https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap says it.

ianmcook: I have no idea, but if the docs say to use `homebrew-tap` then that seems like a good reason to.

ianmcook:
Thanks for doing this!

.goreleaser.yaml
zeroshade: at a minimum we should do Version + Major/Minor

amoeba: Sounds good.

amoeba: Edited my suggestion.

zeroshade:
I created `columnar-tech/homebrew-tap` and a PAT for the workflow to use.

PR #144, opened 2025-10-20
feat(info): add info subcommand

### LINKED ISSUE #82: Mechanism for telling the user what platforms a driver is available
This might be frustrating, maybe not now but if drivers on the CDN aren't packaged for our usual mix of platforms,

```PS
PS C:\Users\Bryce\src\dbc> go run ./cmd/dbc install sqlite
Error: no package found for platform 'windows_arm64'
exit status 1
PS C:\Users\Bryce\src\dbc>
```

Okay, so what platforms _is_ the driver available for? I think we should give the user a way to find this out, either by extended `dbc search` or by pointing the user to a `dbc info` command (#26)

zeroshade:
closes #82

### DIFF
--- a/cmd/dbc/info.go
+++ b/cmd/dbc/info.go
@@ -15,14 +15,10 @@
 package main
 
 import (
-	"os"
 	"strings"
 
-	"github.com/charmbracelet/bubbles/viewport"
 	tea "github.com/charmbracelet/bubbletea"
-	"github.com/charmbracelet/lipgloss"
 	"github.com/columnar-tech/dbc"
-	"github.com/mattn/go-isatty"
 )
 
 type InfoCmd struct {
@@ -46,10 +42,8 @@ func (c InfoCmd) GetModel() tea.Model {
 type infoModel struct {
 	baseModel
 
-	driver   string
-	ready    bool
-	viewport viewport.Model
-	drv      dbc.Driver
+	driver string
+	drv    dbc.Driver
 }
 
 func (m infoModel) Init() tea.Cmd {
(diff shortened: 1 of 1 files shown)


cmd/dbc/info.go
```diff
@@ -0,0 +1,146 @@
...
+		if m.viewport.Width == 0 {
+			m.viewport = viewport.New(msg.Width, msg.Height-marginHeight)
+			m.viewport.YPosition = 0
+		} else {
+			m.viewport.Width = msg.Width
+			m.viewport.Height = msg.Height - marginHeight
+		}
+	case dbc.Driver:
+		if isatty.IsTerminal(os.Stdout.Fd()) {
+			cmds = append(cmds, tea.EnterAltScreen)
```
amoeba: Hrm, I'm seeing info open a pager-like interface, I don't like that as a default behavior. Was this intentional?

cmd/dbc/info.go
```diff
@@ -0,0 +1,146 @@
...
+func formatDriverInfo(drv dbc.Driver) string {
+	info := drv.MaxVersion()
+	var b strings.Builder
+
+	b.WriteString(bold.Render("Driver: ") + nameStyle.Render(drv.Path) + "\n")
+	b.WriteString(bold.Render("Version: ") + info.Version.String() + "\n")
+	b.WriteString(bold.Render("Title: ") + drv.Title + "\n")
+	b.WriteString(bold.Render("License: ") + drv.License + "\n")
+	b.WriteString(bold.Render("Description: ") + drv.Desc + "\n")
+	b.WriteString(bold.Render("Available Packages:") + "\n")
```
amoeba, line 74: Do we want to add URLs to the output? Are we using it yet? May as well even if not.

cmd/dbc/info.go
zeroshade: yes that was intentional. I was matching the behavior of `apt-get info` etc. I can remove that pager interface if desired

cmd/dbc/info.go
zeroshade: the URLs are the locations to download the actual packages, I don't think there's any reason to add those to the output

cmd/dbc/info.go
zeroshade: removed the pager interface and updated the PR

cmd/dbc/info.go
amoeba: Oh, okay. I was mixing that up. Thanks.

cmd/dbc/info.go
amoeba: Yeah, I think if we want do automatic paging on commands, it should be overall paradigm the CLI uses consistently so  it's better to just print normally for now. Thank you!

amoeba (approved):
Thanks!

PR #146, opened 2025-10-21
docs: add homebrew instructions to docs

zeroshade:


README.md
```diff
@@ -132,6 +132,21 @@ architectures:
 - `columnar/dbc:latest`
 - `columnar/dbc:{major}.{minor}.{patch}`, e.g. `columnar/dbc:0.0.1`
 
+## Homebrew
+
+`dbc` is available via the Columnar Homebrew tap. The tap will first need
+to be added to your environment:
+
+```console
```
eitsupi, line 140: Maybe `sh` instead of `console`?

README.md
```diff
@@ -132,6 +132,21 @@ architectures:
...
+$ brew tap columnar-tech/tap
+```
+
+Once you've done this you can install the `dbc` cask:
+
+```console
```
eitsupi, line 146: Ditto.

README.md
amoeba: We need `console` if we're including the prompt character in the code snippet and we're generally including prompt characters so I think this and the below instance can stay `console`.

README.md
```diff
@@ -132,6 +132,21 @@ architectures:
 - `columnar/dbc:latest`
 - `columnar/dbc:{major}.{minor}.{patch}`, e.g. `columnar/dbc:0.0.1`
 
+## Homebrew
+
+`dbc` is available via the Columnar Homebrew tap. The tap will first need
```
amoeba:
```suggestion
dbc is available via the Columnar Homebrew tap. The tap will first need
```

amoeba (approved):
Thanks, just one change.

README.md
zeroshade: why remove the backticks for formatting?

amoeba: Just to be consistent with the style used everywhere else in the docs.

dbc is for referring to the software
`dbc` is for referring to what a user types into code or a terminal

README.md
eitsupi: Thanks, I see your point.
But I suspect that the README is not the part of Mkdocs site, is it?

ianmcook: +1, in the other boxes in the README, we use `sh` so we should do that here.

amoeba: Ah yeah. Sorry @eitsupi, you're absolutely right.

PR #148, opened 2025-10-21
chore(cmd): remove namesonly flag

### LINKED ISSUE #58: Consider removing or renaming --namesonly flag to dbc search
While writing up docs trying to explain --namesonly, I was having a hard time explaining it. Do we really need it right now?

I have two issues with it as it stands:

- It's called "names only" but there's no such thing a driver name (drivers have titles though). If we keep the flag, we need to rename something.
- search, by default, doesn't show titles. And do users really want to filter by the contents of a title?

@zeroshade is there a good reason to have the flag right now versus adding it in later if users tell s they need something like it?

zeroshade:
closes #58

### DIFF
--- a/docs/guides/finding_drivers.md
+++ b/docs/guides/finding_drivers.md
@@ -117,15 +117,3 @@ $ dbc search --verbose
    Available Versions:
     ╰── 0.1.0
 ```
-
-### Searching Names Only
-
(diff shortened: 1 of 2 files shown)


amoeba (approved):
Thanks. Added a commit with docs updates too.

PR #152, opened 2025-10-23
fix(tests): add prog.Wait in test

zeroshade:
This might hopefully fix the flakiness?

### DIFF
--- a/cmd/dbc/subcommand_test.go
+++ b/cmd/dbc/subcommand_test.go
@@ -133,6 +133,7 @@ func (suite *SubcommandTestSuite) runCmd(m tea.Model) string {
 
 	var err error
 	m, err = prog.Run()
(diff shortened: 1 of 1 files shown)


ianmcook:
I'll try re-running the flaky jobs several times here ⌛

ianmcook (approved):
I re-ran the tests four times, and there were no failures. So it seems statistically probable that this fixed it!

amoeba:
Is this flakiness what we had hoped to fix with this? https://github.com/columnar-tech/dbc/actions/runs/18792503522/job/53625842071?pr=154

zeroshade:
Hmm, so we fixed the flakiness on macOS but not on Windows -_- oy

amoeba:
Worth trying to improve things, at the very least.

Issue #153, opened 2025-10-24
Add SECURITY.md


amoeba:

ianmcook:
We might want to do this with a `.github` repo like we do in adbc-drivers. @lidavidm ?

amoeba:
Does that work org-wide and avoid needing to have this file in each repo?

amoeba:
Seems so? https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

ianmcook:
Yes. You can put `CODE_OF_CONDUCT.md` there also and it shows up org-wide in all repos.

ianmcook:
(But you need to delete the copies that exist in the individual repos for the org-wide copies to take effect.)

amoeba:
I filed a PR to remove the COC markdown file in this repo now that our .github is set up: https://github.com/columnar-tech/dbc/pull/155.

PR #154, opened 2025-10-24
docs: add SECURITY.md

amoeba:
Closes https://github.com/columnar-tech/dbc/issues/153


amoeba:
Let's do this org-wide.

PR #156, opened 2025-10-24
docs: set up social image preview cards

### LINKED ISSUE #105: Create social preview image
1200x627px

amoeba:
Enables the mkdocs-material [social plugin](https://squidfunk.github.io/mkdocs-material/plugins/social/) which produces a reasonable social media preview image for all pages. There's some customization ability we can follow up with where needed.

Outside of CI, this requires some setup so I've documented that. On GitHub Actions, the necessary dependencies should be available.

Also enables caching on GHA.

Closes #105

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -21,6 +21,10 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r docs/requirements.txt
+      - uses: actions/cache@v4
+        with:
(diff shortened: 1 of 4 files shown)


amoeba:
Without a good PR deploy preview, the only way to really test this is locally. When you build the site, you can see the cards in site/assets/social. The cards take on the site's theme automatically. Here's an example:

driver_manifest

amoeba:
Note for myself that we'll need to watch CI to see if the new cache step seems to work.

Issue #149, opened 2025-10-22
Add uninstallation instructions


ianmcook:
On the Installation docs page, we should show how to remove dbc from your system (which might vary depending on how you installed it). The uv docs does this nicely: https://docs.astral.sh/uv/getting-started/installation/#uninstallation

PR #157, opened 2025-10-24
docs: add uninstallation section to install docs

amoeba:
Closes https://github.com/columnar-tech/dbc/issues/149

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -194,19 +194,19 @@ dbc can generate shell completions for a number of common shells.
 
 To remove dbc from your system, run the uninstall command corresponding to your installation method.
 
-=== "macOS and Linux command-line installer"
+=== "macOS / Linux (CLI)"
 
     ```console
     $ rm $HOME/.local/bin/dbc
     ```
 
-=== "Windows command-line installer"
+=== "Windows (CLI)"
 
     ```console
     $ powershell.exe -c "rm $HOME/.local/bin/dbc.exe"
     ```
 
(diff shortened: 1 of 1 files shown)


amoeba:
Hrm, I wasn't sure about having explicit uninstall commands for each installation method like that and kinda wanted to just document uninstallation for methods where users might not be expected to know how to uninstall things.

And to a lesser extent, I think we want to avoid having too many tabs if we can since the UX is so poor on it (it's very hard to notice extra tabs and even harder to switch to them). 1-3 tabs with short titles is fine.

Thoughts on my above?

ianmcook:
> Hrm, I wasn't sure about having explicit uninstall commands for each installation method like that and kinda wanted to just document uninstallation for methods where users might not be expected to know how to uninstall things.

If creating and maintaining this list of uninstall instructions to cover all the documented installation methods were going to be a big challenge for us, I'd be more inclined to agree. But this is easy. So I think it's worthwhile to include them. It seems a bit user-hostile for us to say "you're own your own" if you happened to install with one of the other methods but you don't know how to uninstall with that method.

> And to a lesser extent, I think we want to avoid having too many tabs if we can since the UX is so poor on it (it's very hard to notice extra tabs and even harder to switch to them). 1-3 tabs with short titles is fine.

Agreed on this; my titles were way too long. I shortened them in e76e797 so they fit without horizontal scrolling on desktop now.

amoeba:
That's totally fair that it's easy for us to add (and keep in sync with our install methods).

Thanks for shortening the titles, that looks good to me on desktop.

PR #161, opened 2025-10-25
docs: fix social image preview card on docs index page

amoeba:
This replaces the big text "Introduction" on our index page's social image preview with "Documentation". The other SEO tags for the page are unchanged as far as I can tell.


ianmcook:
This also changes the title in the top bar, but I think it's fine.
image

Issue #101, opened 2025-09-26
Link to quickstarts from README and docs


ianmcook:
I think we're eventually going to pull the content from https://github.com/columnar-tech/adbc-quickstarts into the dbc docs, but in the meantime, we should probably link to that repo from the README and docs.

amoeba:
Closed by https://github.com/columnar-tech/dbc/pull/162.

amoeba:
I'll put up a PR to add a link to the README too in a sec.

PR #162, opened 2025-10-25
docs: list more installation options on index page

ianmcook:
Any reason we shouldn't add the other installation options to the Installation section of the homepage?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -472,3 +472,9 @@ dbc is a command-line tool for installing and managing [ADBC](https://arrow.apac
 - Cross-platform: Runs on macOS, Linux, and Windows
 - Installable with pip, Docker, and more (See [Installation](./getting_started/installation.md))
 - Works great in CI/CD environments
+
+## Help
+
+- Join the [Columnar Community Slack](https://join.slack.com/t/columnar-community/shared_invite/zt-3gt5cb69i-KRjJj~mjUZv5doVmpcVa4w)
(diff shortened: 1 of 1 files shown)


ianmcook:
Alternatively, we could move the installation instructions to the top, like in this mockup:
[Documentation - dbc.pdf](https://github.com/user-attachments/files/23145092/Documentation.-.dbc.pdf)

Thoughts on whether this is better?

amoeba:
I'm liking the mockup with the install at the top. My originally thought on this page was to show just one thing for "what does dbc do" and to hide the rest of the info elsewhere. But I don't think that model works well for dbc whereas it works better for a tool like uv. If we can't show the one thing dbc does, we should show all three steps in line like you have.

Let me try it out.

amoeba:
I pushed up my changes, this is ready for a review @ianmcook.

zeroshade (approved):

ianmcook:
This looks great, thanks @amoeba!

The steps 1, 2, 3 makes me wonder if we should bring back the "easy as 1-2-3" tagline 🤔 

The only thing I don't like about this is that the Features list is now at the bottom of the index page, and it has a lack-of-closure vibe when you get to the end. How about adding a **Help** section below that that suggests ways to get help:
- Join the Columnar Community Slack (with a link)
- Open an issue or start a discussion in the dbc repo (with a link)
- See the ADBC Quickstarts (with a link) <-- that also addresses #101

amoeba:
How does that look? I added a segue sentence before the 1/2/3 and added the Help section as you suggested.

PR #167, opened 2025-10-27
fix: support user-level install in Windows MSI

### LINKED ISSUE #165: support per-user installation on Windows
The current dbc Windows MSI installs to Program Files which requires elevated permissions. On a non-admin user account, the user is required to know the admin password to install dbc.

The winget platform seems to require installers install cleanly to a non-elevated directory, https://learn.microsoft.com/en-us/windows/package-manager/package/repository#submission-expectations,

> The application installs and uninstalls correctly for both administrators and non-administrators.

I think this is a general trend on the Windows ecosystem as more and more Windows programs I use install to $HOME/AppData/Local or somesuch. 

For the dbc Windows installer, I think we might want to default to a per-user install with a system-wide install as a non-default option.

amoeba:
Closes #165 

This modifies our Wix setup to support both user-level and system-wide installs. It does this by adding some variables and a new dialog where the user picks a user- or system-level install first before installing.

### DIFF
--- a/resources/dbc.wxs
+++ b/resources/dbc.wxs
@@ -26,12 +26,31 @@ limitations under the License.
   <?define ProgramFilesFolder = "ProgramFilesFolder" ?>
   {{ end }}
   <Fragment>
-    <UI Id="WixUI_Minimal">
+    <!-- Custom actions to set installation folders -->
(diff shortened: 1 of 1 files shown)


zeroshade:
Closing in favor of #168

PR #185, opened 2025-11-04
docs: add cpp driver manager instructions

ianmcook:
Adds C++ driver manager instructions to the main docs page.

### DIFF
--- a/docs/guides/driver_manager.md
+++ b/docs/guides/driver_manager.md
@@ -77,7 +77,7 @@ The best place to find detailed information on driver manager installation is al
     === "conda"
 
         ```console
-        $ conda install -c conda-forge adbc-driver-manager
+        $ conda install -c conda-forge libadbc-driver-manager
         ```
 
     === "apt"


docs/guides/driver_manager.md
```diff
@@ -77,7 +77,7 @@ The best place to find detailed information on driver manager installation is al
     === "conda"
 
         ```console
-        $ conda install -c conda-forge adbc-driver-manager
+        $ conda install -c conda-forge libadbc-driver-manager
```
ianmcook, line 80: @amoeba I think:
https://anaconda.org/conda-forge/adbc-driver-manager <-- this is the Python one
https://anaconda.org/conda-forge/libadbc-driver-manager <-- this is the C++ one
The former might work because it contains the C++ library, but I think we want the latter.

amoeba: Yep.

docs/index.md
```diff
@@ -69,6 +69,14 @@ dbc is a command-line tool for installing and managing [ADBC](https://arrow.apac
 
     <br/>3. [Install a driver manager](./guides/driver_manager.md) and load drivers in any supported language:
 
+    === "C++"
+
+        ```cpp
+        #include <arrow-adbc/adbc.h>
+
+        AdbcDatabaseSetOption(&database, "driver", "bigquery", &error)
```
amoeba, line 77: Do we want to go a bit further? Technically, our examples in different languages don't all do the exact same thing. In this case, we're not actually loading the driver or even causing the driver to be loaded. Same for R actually AFAIK.

This doesn't have the same conciseness though:

```c++
AdbcError error = {};
AdbcDatabase database = {};

CHECK_ADBC(AdbcDatabaseNew(&database, &error));
CHECK_ADBC(AdbcDatabaseSetOption(&database, "driver", "duckdb", &error));
CHECK_ADBC(AdbcDatabaseSetOption(&database, "path", "games.duckdb", &error));
CHECK_ADBC(AdbcDriverManagerDatabaseSetLoadFlags(
    &database, ADBC_LOAD_FLAG_DEFAULT, &error));
CHECK_ADBC(AdbcDatabaseInit(&database, &error));
```

ianmcook: The goal of these code snippets in the different languages is to show:
- Which library you'll need to load to use the driver (or if there are multiple, then just the most important one).
- Which line of code specifies the driver name.

As much as possible, I'd like to keep them all like that, for maximum conciseness. We should direct people to the quickstarts to see full examples.

amoeba: Works for me. I wanted to tweak the language above,

> "3. [Install a driver manager](./guides/driver_manager.md) and load drivers in any supported language:"

to be more accurate but couldn't not make it awkward. I think the imprecision is fine.

amoeba (approved):

Issue #176, opened 2025-10-29
add dbc docs subcommand for opening and/or printing driver docs


amoeba:
The current set of subcommands let users install drivers but every driver requires at least some amount of documentation for setup. We could add a new "docs" subcommand which,

1. When run as `dbc docs`, open the main dbc docs
2. When run as `dbc docs <DRIVER>`, opens the documentation for the driver

For (2), I'd like that to pull from the driver index rather than hard-coding where to get the docs into dbc itself. This would require an update to the CDN and packaging process.

PR #186, opened 2025-11-05
feat(docs): add subcommand for opening documentation

amoeba:
Adds the implementation and documentation for a new subcommand, `docs`.

If you run `dbc docs` with no driver, the main dbc docs site opens in your default browser. If you run `dbc docs DRIVER`, we look for that driver in the driver registry (the CDN). If not found, we error. If found, we look in a list of temporary, fallback URLs for a docs URL and open that. For drivers that exist in the registry but have no fallback URL, we error.

Closes https://github.com/columnar-tech/dbc/issues/176

### DIFF
--- a/cmd/dbc/docs.go
+++ b/cmd/dbc/docs.go
@@ -26,7 +26,7 @@ var dbcDocsUrl = "https://docs.columnar.tech/dbc/"
 
 // Support drivers without a docs URL defined in the index
 var fallbackDriverDocsUrl = map[string]string{
-	"duckdb":     "https://arrow.apache.org/adbc/current/driver/duckdb.html",
+	"duckdb":     "https://duckdb.org/docs/stable/clients/adbc",
 	"flightsql":  "https://arrow.apache.org/adbc/current/driver/flight_sql.html",
 	"postgresql": "https://arrow.apache.org/adbc/current/driver/postgresql.html",
 	"snowflake":  "https://arrow.apache.org/adbc/current/driver/snowflake.html",


amoeba:
Tagged for reviews. I wasn't sure if I liked `dbc doc` or `dbc docs` better. Let me know.

ianmcook:
Thanks for doing this @amoeba. I prefer `dbc docs`.

amoeba:
Changed from "doc" to "docs" in daadd80305d6a8f06d647f62efd0c1db11bfee9b. 

@ianmcook do you want to take it for a test drive and @zeroshade do you want to check my implementation (code and tests)?

ianmcook:
Thanks @amoeba, this is great.

I'm wondering if we really need the `y/n` prompt. Does the courtesy of this outweigh the inconvenience? I'm aware of at least one other command-line tool that has an identical `docs` subcommand that does not prompt the user for permission before opening the browser: `npm docs <package>`. However it also has a `--no-browser` option that prints a message like this instead of opening the browser:

```
<package> docs available at the following URL:
https://example.com
```

amoeba:
I wasn't sure which I preferred and found I didn't have a lot of prior art to pull from. Another example is R in some cases (like Windows) and it doesn't prompt IIRC. The only other examples I could think of were some of the CLIs that do auth and those prompt but usually they are giving you instructions and sometimes a code to copy paste.

I'll rework this to not prompt and add a `--no-open` or similar flag.

amoeba:
This is all refactored now. It now just opens the URL without prompting or printing anything to the terminal. And it also supports a `--no-open` flag that works like this,

```console
$ go run ./cmd/dbc docs --no-open
dbc docs are available at the following URL:
https://docs.columnar.tech/dbc/

~/src/columnar-tech/dbc feat/subcommand--doc
$ go run ./cmd/dbc docs sqlite --no-open
sqlite driver docs are available at the following URL:
https://arrow.apache.org/adbc/current/driver/sqlite.html

~/src/columnar-tech/dbc feat/subcommand--doc
$ go run ./cmd/dbc docs bigquery --no-open
Error: no documentation available for driver `bigquery`
exit status 1
```

I went with `--no-open` since we're already using `--no-foo` in at least one other place.

amoeba:
CI seems good, I manually tested on Windows just to be sure the package we're using is as cross-platform as they say they are. It seems to work :)

Ready for a final review @ianmcook

cmd/dbc/docs.go
```diff
@@ -0,0 +1,149 @@
...
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/cli/browser"
+	"github.com/columnar-tech/dbc"
+)
+
+var dbcDocsUrl = "https://docs.columnar.tech/dbc/"
+
+// Support drivers without a docs URL defined in the index
+var fallbackDriverDocsUrl = map[string]string{
+	"duckdb":     "https://arrow.apache.org/adbc/current/driver/duckdb.html",
```
ianmcook: I wonder if we should link to `https://duckdb.org/docs/stable/clients/adbc` instead?

amoeba: Hrm, yeah we should. And we should update the DuckDB page in the ADBC docs a bit too (I'll do that).

amoeba: Changed in 849ee577c3d1309d607ce97be1302a3321152f4d.

ianmcook:
Since this is a new subcommand, do we need to change anything in the shell completions?

amoeba:
Good call. We should do that during release. I'll go make sure that's in our checklist.

PR #188, opened 2025-11-05
docs: document winget install option

ianmcook:
This documents how to install and uninstall dbc with WinGet. Thanks @eitsupi for making this possible!

This also improves the naming of the install options a bit, and fixes an mkdocs build warning.

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -122,6 +122,14 @@ You can also download the latest installer using the following URL:
 |--------------|---------------------------------------------------------|
 | `x64`        | <https://dbc.columnar.tech/latest/dbc-latest-x64.msi>   |
 
+## WinGet
+
+On Windows, you can install dbc using [WinGet](https://learn.microsoft.com/en-us/windows/package-manager/winget/):
+
+```console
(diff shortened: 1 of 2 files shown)


docs/getting_started/installation.md
```diff
@@ -122,6 +122,14 @@ You can also download the latest installer using the following URL:
 |--------------|---------------------------------------------------------|
 | `x64`        | <https://dbc.columnar.tech/latest/dbc-latest-x64.msi>   |
 
+## WinGet
+
+On Windows, you can install dbc using [WinGet](https://learn.microsoft.com/en-us/windows/package-manager/winget/):
+
+```console
+$ winget install dbc
+```
```
eitsupi, line 131: In fact, installing with winget is the same as using the MSI installer, so it might be a good idea to make it a subsection of MSI.

ianmcook: They're the same under the hood, but from the dbc user's perspective, the MSI and WinGet options are distinct. I think it makes sense to keep this in its own section like this.

eitsupi: Sounds good, thanks!

PR #189, opened 2025-11-07
docs: replace pipx with uv

esadek:
Replace pipx with uv in the PyPI installation section of the README.
uv is more popular and faster than pipx.

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -122,6 +122,14 @@ You can also download the latest installer using the following URL:
 |--------------|---------------------------------------------------------|
 | `x64`        | <https://dbc.columnar.tech/latest/dbc-latest-x64.msi>   |
 
+## WinGet
+
+On Windows, you can install dbc using [WinGet](https://learn.microsoft.com/en-us/windows/package-manager/winget/):
+
+```console
+$ winget install dbc
+```
+
 ## Docker
 
 We publish [Docker](https://docker.io) images for each dbc release.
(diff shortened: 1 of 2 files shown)


amoeba:
Thanks @esadek. Are you seeing projects drop pipx entirely for uv? We're big fans of uv but also don't want to alienate people. My sense is that uv adoption is pretty high for folks on the bleeding edge but maybe not so much more broadly.

esadek:
@amoeba

Homebrew analytics shows that uv has significantly more install events than pipx. Anecdotally, many developers I’ve spoken with have switched from tools like pipx, pyenv, and poetry to uv, and I’ve noticed more projects recommending uv for installation.

In my opinion, if the README suggests a tool for installation from PyPI, it should be uv.
The documentation site still includes pipx instructions. This change is only for the README.

eitsupi:
IMO, pipx is maintained by pypa, the version is > 1.0.0, and it is quite stable, so I don't think there is any need to replace to uv, which is a version lower than 1.0.0.

Also, users currently using uv may be early adopters, so it may be easier to look at a pipx command and replace it with the equivalent uv command than the other way around.

esadek:
> Also, users currently using uv may be early adopters, so it may be easier to look at a pipx command and replace it with the equivalent uv command than the other way around.

That's fair. Most uv tool users should know `pipx install dbc` can be replaced with `uv tool install dbc` and `pipx run dbc` can be replaced with `uvx dbc`.

amoeba:
Thanks for your thoughts @eitsupi.

@esadek how about we keep the readme as-is for now and closing this PR? I appreciate you taking the time to help align our docs with user expectations. We can revisit a change like this in the future.

Issue #116, opened 2025-09-30
Document how to use dbc in Python notebooks


ianmcook:
Show how to use it with Jupyter, Colab, etc. Use `!dbc install`.

esadek:
I'd be happy to take up this issue if it isn't being worked on.

Here's a basic example with DuckDB which does not require an existing database file:

```python
!dbc install duckdb
```

```python
%pip install adbc_driver_manager pyarrow
```

```python
from adbc_driver_manager import dbapi
```

```python
with (
    dbapi.connect(driver="duckdb", db_kwargs={"path": "penguins.duckdb"}) as con,
    con.cursor() as cursor,
):
    cursor.execute("CREATE TABLE IF NOT EXISTS penguins AS FROM 'http://blobs.duckdb.org/data/penguins.csv'")
    cursor.execute("SELECT * FROM penguins")
    table = cursor.fetch_arrow_table()
```

```python
print(table)
```

PR #190, opened 2025-11-07
docs: add python notebooks guide

esadek:
Add a guide for installing and using dbc in Python notebooks.

The included code has been tested in both a local notebook and Google Colab.

Closes https://github.com/columnar-tech/dbc/issues/116


docs/guides/python_notebooks.md
```diff
@@ -0,0 +1,61 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Python Notebooks
+
+dbc can be installed and used directly in Python notebooks (such as Jupyter or Google Colab).
```
amoeba:
```suggestion
dbc can be installed and used directly in Python notebooks (such as [Jupyter](https://jupyter.org) or [Google Colab](https://colab.google)).
```

docs/guides/python_notebooks.md
```diff
@@ -0,0 +1,61 @@
...
+
+```python
+from adbc_driver_manager import dbapi
+```
+
+Connect to a database via ADBC, create a cursor, execute queries, and fetch the result as a PyArrow Table:
+
+```python
+with (
+    dbapi.connect(driver="duckdb", db_kwargs={"path": "penguins.duckdb"}) as con,
```
amoeba:
```suggestion
    dbapi.connect(driver="duckdb") as con,
```

How about just using an in-memory database for the example here? It reduces the amount of code and it's not a key detail.

docs/guides/python_notebooks.md
```diff
@@ -0,0 +1,61 @@
...
+    con.cursor() as cursor,
+):
+    cursor.execute("CREATE TABLE IF NOT EXISTS penguins AS FROM 'http://blobs.duckdb.org/data/penguins.csv'")
```
amoeba:
```suggestion
    cursor.execute("CREATE TABLE IF NOT EXISTS penguins AS FROM 'https://blobs.duckdb.org/data/penguins.csv'")
```

docs/guides/python_notebooks.md
```diff
@@ -0,0 +1,61 @@
...
+    cursor.execute("SELECT * FROM penguins")
+    table = cursor.fetch_arrow_table()
+```
+
+Print the table:
+
+```python
+print(table)
```
amoeba, line 60: How about adding the output? It may be helpful if users aren't that familiar with PyArrow or the dataset.

docs/guides/python_notebooks.md
eitsupi: Considering something like marimo, I wonder if it's appropriate to use the term "Python Notebook" to refer only to Jupyter / ipynb.
(IIUC, Colab is also Jupyter.)

So, I think this page should actually be called Jupyter instead of Python Notebook.

ianmcook: Or even better: @esadek, could you could test dbc in a Marimo notebook and include instructions for how to use it there?

eitsupi: On marimo, we should use something like `import subprocess; subprocess.run(["dbc", "install", "duckdb"])`
https://docs.marimo.io/guides/coming_from/jupyter/#magic-commands

amoeba: Yeah, when reviewing this PR I looked at and tested marimo and I don't know if it makes sense to do more than add a note about it. I doubt we want to tell users to run dbc like the above since it's not very convenient.

ianmcook: Yep, I think a note like this would be sufficient:

> If you're using a Python notebook that doesn't support magics (`%`) or shell escapes (`!`) then use `subprocess.run` to run dbc commands from your Python code.

esadek: After testing marimo, I've concluded that a dedicated guide or section would be best.

In marimo, packages are typically managed via the terminal (external or internal), inline script metadata, or the integrated "Manage packages" panel. This differs from Jupyter or Google Colab, where `!pip install` is common.

Drivers can be installed with dbc via the terminal (external or internal) or in a cell using `subprocess.run`.

Locally, `dbapi.connect(driver="duckdb")` works as expected. However, on molab (marimo’s cloud platform), `adbc_driver_manager` throws an error. Instead, the full driver path must be used: `dbapi.connect(driver="/tmp/uv-venv/etc/adbc/drivers/duckdb")`.

ianmcook: Thanks for looking into this @esadek. Could you open a separate issue for us to look into the molab problem? Thanks.

amoeba: Issue for this thread: https://github.com/columnar-tech/dbc/issues/201

PR #194, opened 2025-11-10
chore: add path filters to workflows

esadek:
Add path filters to the Go and Integration workflows to skip runs for unrelated changes (e.g., documentation updates).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -22,13 +22,14 @@ on:
       - '**.go'
       - 'go.mod'
       - 'go.sum'
+      - 'cmd/dbc/**'
       - 'ci/scripts/run_integration.sh'
   pull_request:
     paths:
       - '**.go'
-      - 'cmd/dbc/**'
       - 'go.mod'
       - 'go.sum'
+      - 'cmd/dbc/**'
       - 'ci/scripts/run_integration.sh'
 
 concurrency:


esadek:
> An alternative would be to just ignore (paths-ignore) the docs folder. We can test this out though.

My initial idea was to use `paths-ignore`, but there are many more files and folders to exclude beyond just `docs/`.

.github/workflows/go.yml
```diff
@@ -17,7 +17,17 @@ on:
   push:
     branches:
       - 'main'
+    paths:
```
zeroshade, line 20: should we also include `.goreleaser.yaml` and `columnar.pubkey` on the lists?

.github/workflows/integration.yml
```diff
@@ -18,7 +18,17 @@ on:
   push:
     branches:
       - "main"
+    paths:
+      - '**.go'
```
zeroshade, line 22: shouldn't these also have `cmd/dbc/**`?

amoeba: Yeah I think so, I'll add this.

.github/workflows/go.yml
amoeba: I don't think so? The build this workflow triggers technically reads in the pubkey (to embed) but goreleaser is only used by the release workflow (which is gated by other means).

.github/workflows/integration.yml
amoeba: Added in 02dd73e9fccfa41086d3a564885af0d7b768e6b8.

amoeba:
I'm going to merge this and leave https://github.com/columnar-tech/dbc/pull/194#discussion_r2529135779 as an option to follow up on if we feel we need it.

PR #200, opened 2025-11-11
docs: fix icon paths

esadek:
Fix documentation icon paths.

### DIFF
--- a/docs/manifest.webmanifest
+++ b/docs/manifest.webmanifest
@@ -1,12 +1,12 @@
 {
   "name": "dbc Documentation",
   "short_name": "dbc Docs",
-  "start_url": "/dbc/",
-  "scope": "/dbc/",
+  "start_url": "https://docs.columnar.tech/dbc/",
+  "scope": "https://docs.columnar.tech/dbc/",
   "icons": [
-    { "src": "assets/icon-192.png", "type": "image/png", "sizes": "192x192" },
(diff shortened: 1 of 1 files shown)


docs/overrides/main.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 
 {% block extrahead %}
-  <link rel="icon" href="assets/icon.svg" type="image/svg+xml">
-  <link rel="apple-touch-icon" href="assets/apple-touch-icon.png">
-  <link rel="manifest" href="manifest.webmanifest">
+  <link rel="icon" href="/dbc/assets/icon.svg" type="image/svg+xml">
```
amoeba: Does this work with mkdocs-material?

```suggestion
  <link rel="icon" href="{{base_url}}/assets/icon.svg" type="image/svg+xml">
```

What you have here should work but using the variable is safer/better.

ianmcook:
I think we should also make the paths inside `manifest.webmanifest` into absolute paths.

docs/overrides/main.html
esadek: Yes, this works. Thanks

ianmcook:
Thanks @esadek, this works great. Bookmarks in Safari look beautiful, and when you do "Add to Home Screen" on iOS, dbc Docs looks like a proper app.

PR #205, opened 2025-11-13
chore: add editorconfig file

esadek:
Add a `.editorconfig` file to maintain consistent coding styles across developers and editors.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,4 @@
+# Code formatting fixes
+1edfbb5a86f4f3e3d00e34ea7c2e034fee6a217f
+7b8d457a991b6cc0c47e4386f0881cad1183aaee
+e89b95ccc44c73ceaaacb1fa2393f13b962bfcc2


amoeba:
Thanks for this. Something along these lines would be good. I have my editor set up to normalize all the things I see in the editorconfig but I can tell @zeroshade uses some other method. What I do now is save without format-on-save. @zeroshade would this work okay for your editor setup?

btw I ran `eclint fix` on the repo with the above editorconfig and it breaks formatting in ways we don't want. But I think it'd be good to tidy up the repo in this PR if we can. Do you want to see about doing that @esadek and using git `ignoreRevsFile`?

esadek:
Thank you for checking with `eclint`. I wasn’t aware of EditorConfig validation tools like this.

I’ve made formatting fixes, added commit hashes to `.git-blame-ignore-revs`, and rerun `editorconfig-checker`. The only remaining errors are "Wrong indentation type (spaces instead of tabs)" for `bash.go`, `fish.go`, and `zsh.go` in `cmd/dbc/completions/`. This is intentional, as space indentation is used inside multiline strings.

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{go,mod}]
+indent_style = tab
```
zeroshade, line 11: why switch to indent_style tab for go instead of leaving it as space?

esadek: All of the go files in the repo use tabs for indentation. This also seems to be the standard, as gofmt uses tabs.

zeroshade: Okay, fair enough 😄

PR #206, opened 2025-11-13
docs: add pixi environment

esadek:
Add a docs environment, dependencies, and tasks to the Pixi manifest. With Pixi, there’s no need to manually create and activate a Python virtual environment or install mkdocs and cairo.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7
         with:
           pixi-version: v0.59.0
           cache: true


pixi.toml
amoeba, line 22: To be uniform, should we use an environment for the cdn (cdn-generate, cdn-serve) stuff too?

amoeba (commented):
Thanks. Added one question as a comment above.

I think we might also consider updating everywhere we're using venvs to just use the new pixi setup (docs/ folder, CI). What do you think?

esadek:
> I think we might also consider updating everywhere we're using venvs to just use the new pixi setup (docs/ folder, CI). What do you think?

I agree. Would you like that included in this PR?

amoeba:
Either way. Doing it in a follow-up is fine too.

esadek:
I've updated `docs/README.md` and `.github/workflows/publish_docs.yml` to use Pixi.

.github/workflows/publish_docs.yml
```diff
@@ -28,19 +28,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v4
+      - uses: prefix-dev/setup-pixi@v0.9.3
```
amoeba:
```suggestion
      - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7
```

Small safety thing: for third-party (i.e., non GitHub) actions workflows, we like to use the commit hash to reference workflows instead of a tag since the tag can change.

esadek:
Yes, having Cairo automatically installed into the environment is awesome.

amoeba (approved):

Issue #203, opened 2025-11-12
`dbc search -v` does not print paths to installed drivers on Windows


ianmcook:
```sh
C:\Users\cooki>dbc search duckdb -v
• duckdb
   Title: DuckDB Driver
   Description: An ADBC driver for DuckDB developed by the DuckDB Foundation
   License: MIT
   Installed Versions:
    ╰── 1.4.2
        ╰── user =>
   Available Versions:
    ├── 1.4.0
    ├── 1.4.1
    ╰── 1.4.2
```

(Notice nothing after `user =>`.)

amoeba:
I can confirm this is happening and see why. On non-Windows platforms, we normally show the root dir for the config level in the output. On Windows, we install to the registry for user/system levels so I wonder we should show the registry key, the registry key and the driver path, or just the diver path.

PR #207, opened 2025-11-13
fix(search,windows): show installed location for all config levels

amoeba:
Fixes an issue where the driver install location wasn't showing up when a user runs `dbc search -v` on Windows. This was only impacting user and system level installs and the reason we didn't catch this is that we don't test those levels in CI. 

This PR updates the Windows code paths that set the FileInfo member on the DriverInfo struct so it's set to the registry root for the appropriate level. We just weren't setting it before for registry user and system levels. I went with the registry key root instead of the location of the DLL because the driver is really installed to the registry and we don't want users thinking they can uninstall a driver from %APPDATA% or something. 

Here's an example of what this looks like on my system:

```
• sqlite
   Title: ASF SQLite Driver
   Description: An ADBC driver for SQLite developed under the Apache Software Foundation
   License: Apache-2.0
   Installed Versions:
    ╰── 1.8.0
        ├── system => HKEY_LOCAL_MACHINE\SOFTWARE\ADBC\Drivers
        ├── env => C:\Users\Bryce\src\columnar-tech\dbc
        ╰── user => HKEY_CURRENT_USER\SOFTWARE\ADBC\Drivers
   Available Versions:
```

Closes https://github.com/columnar-tech/dbc/issues/203


config/dirs_windows.go
```diff
@@ -47,6 +47,17 @@ func (c ConfigLevel) key() registry.Key {
 	}
 }
 
+func (c ConfigLevel) rootKeyString() string {
+	switch c {
+	case ConfigUser:
+		return "HKEY_CURRENT_USER"
```
zeroshade: Could we shrink this a bit by using "HKCU" and "HKLM" instead of the full strings?

config/dirs_windows.go
```diff
@@ -143,6 +154,10 @@ func driverInfoFromKey(k registry.Key, driverName string) (di DriverInfo, err er
 	di.Driver.Shared.defaultPath = keyMust(dkey, "driver")
 	di.Driver.Entrypoint = keyOptional(dkey, "entrypoint")
 
+	// For drivers in the registry, set FilePath to the registry key instead
+	// of the filesystem path since that's technically where the driver exists.
+	di.FilePath = fmt.Sprintf("%s\\%s", lvl.rootKeyString(), regKeyADBC)
```
zeroshade, line 159: Since we have k and dkey, can't we get the full key names from those instead of needing the new function?

zeroshade:
> This was only impacting user and system level installs and the reason we didn't catch this is that we don't test those levels in CI.

Should we add tests for CI for this case?

config/dirs_windows.go
amoeba: Yeah good call. I'll do that.

config/dirs_windows.go
amoeba: I tried and failed to get a name from the k (which appears to be just a handle not a rich object). I'll try again though since it'd be nice.

amoeba:
> > This was only impacting user and system level installs and the reason we didn't catch this is that we don't test those levels in CI.
> 
> Should we add tests for CI for this case?

Yeah we should. I'll resurrect the PR for testing user and system, IIRC it's basically ready.

config/dirs_windows.go
zeroshade: Huh, you're right. That really sucks, there's no way to get the current name of the key from the handle.

amoeba: I looked again and I don't see how. It looks like keys are comparable so we can at least simplify it a bit. I'd be happy if you took a look at the package too just in case I'm missing something.

amoeba: See 106e5c3647628c1f7e1f46d76233dfe778ea06e6

config/dirs_windows.go
amoeba: done in 106e5c3647628c1f7e1f46d76233dfe778ea06e6

config/dirs_windows.go
amoeba: hrm, maybe not. the new change results in this,

```
   Installed Versions:
    ╰── 1.8.0
        ├── user => UNK\SOFTWARE\ADBC\Drivers
        ├── system => UNK\SOFTWARE\ADBC\Drivers
        ╰── env => C:\Users\Bryce\src\columnar-tech\dbc
```

amoeba: Actually, my first attempt I was comparing the wrong handles. But comparing the right handles would require changing the function signature of driverInfoFromKey  or moving setting FileInfo out of driverInfoFromKey. So driverInfoFromKey unfortunately has to be modified.

I like what I have now enough. Take another look?

amoeba:
Latest output 

```
• sqlite
   Title: ASF SQLite Driver
   Description: An ADBC driver for SQLite developed under the Apache Software Foundation
   License: Apache-2.0
   Installed Versions:
    ╰── 1.8.0
        ├── user => HKCU\SOFTWARE\ADBC\Drivers
        ╰── system => HKLM\SOFTWARE\ADBC\Drivers
```

amoeba:
Let me file another PR to fix the test failure.

amoeba:
Filed as https://github.com/columnar-tech/dbc/pull/212. We can merge that first if we want.

amoeba:
Rebased.

config/dirs_windows.go
zeroshade: is there any way we can add a test for this? otherwise, this looks good to me

Issue #202, opened 2025-11-12
Add a `--quiet` flag into the `install` command


karakanb:
I want to embed dbc into [Bruin CLI](https://github.com/bruin-data/bruin), for that I'd like to have a `--quiet` flag or sth similar so that dbc does not print any output. I'd like dbc to be quiet to not change anything in our output structure.

PR #208, opened 2025-11-14
feat: add --quiet flag to cli for suppressing output

amoeba:
Adds a global CLI flag, `--quiet` (`-q` for short) that, when set, completely suppresses output (i.e., stdout). Also includes updates to keep the docs in sync.

Command line tools use a variety of methods to enable/disable or enhance/reduce output and there's not really a clear  best way to do this. The [GNU standard](https://www.gnu.org/prep/standards/standards.html#Option-Table) suggests `--quiet` and for the flag to suppress "usual" output. curl takes a different approach with `--silent` which means "print just my final results, not any progress". [uv uses `-q`](https://docs.astral.sh/uv/reference/cli) to reduce output and `-qq` to completely disable output.

This does not include tests because I think it would add quite a bit of noise for not much gain.

Closes https://github.com/columnar-tech/dbc/issues/202

### DIFF
--- a/cmd/dbc/completions/dbc.bash
+++ b/cmd/dbc/completions/dbc.bash
@@ -5,7 +5,7 @@ _dbc() {
     _init_completion || return
 
     local subcommands="install uninstall init add sync search info remove completion"
-    local global_opts="--help -h --version"
+    local global_opts="--help -h --version --quiet -q"
 
     # If we're completing the first argument (subcommand)
     if [[ $cword -eq 1 ]]; then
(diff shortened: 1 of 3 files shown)


amoeba:
For reviewers, if we wanted to go the uv route with `-q`/`-qq`,

- `-q` could suppress bubbletea output and only print what we call "final" output
- `-qq` could then suppress all output, bubbletea and final

zeroshade (changes requested):
LGTM just needs the addition to the shell completion scripts

amoeba:
I sync'd completions and tested fish locally.

Re: the -q vs. -qq idea above, I don't feel we really need two levels of quiet right now so I'm going to leave this PR with just --quiet/-q.

PR #211, opened 2025-11-15
fix: rename old refs to driver list to use the current "driver registry" term

amoeba:
Back before we had formalized names for things, we were calling two different things "driver list". One of those was referring to what we now call the "driver registry" (i.e., the CDN). This PR renames those references.


amoeba:
Rebased.

PR #217, opened 2025-11-23
feat(auth): initial implementation of auth subcommand

zeroshade:
Implements `dbc auth login` and `dbc auth logout` and enables the `makereq` function to utilize any available credentials if the index matches a credential.

Does not yet implement any utilization of encryption or using OS credential stores.

Currently only implements Oauth2 Device Flow and api-key implementation (expecting a `/login` endpoint for api-key to get a token)

### DIFF
--- a/auth/oauth.go
+++ b/auth/oauth.go
@@ -60,7 +60,7 @@ func fetch[T any](u *url.URL, dest *T) error {
 }
 
 func GetOpenIDConfig(issuer *url.URL) (config OpenIDConfig, err error) {
-	for _, p := range []string{"oauth-authorization-server", "openid-configuration"} {
+	for _, p := range []string{"openid-configuration", "oauth-authorization-server"} {
 		u, _ := issuer.Parse("/.well-known/" + p)
 
 		err = fetch(u, &config)


eitsupi:
I played around with it a bit.
I found a few things that might be bugs:

1. `dbc auth` cause panic.
2. `dbc auth login` -> ctrl + c shows `Authentication successful!`

zeroshade:
Thanks for taking a look @eitsupi i'll poke at those

auth/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+	credPath          string
+)
+
+func init() {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic("failed to get user home directory: " + err.Error())
+	}
+
+	credPath = filepath.Join(homeDir, ".local/share/dbc/credentials/credentials.toml")
```
amoeba: Do we want to do this path XDG-style?

auth/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+	if err := LoadCredentials(); err != nil {
+		return err
+	}
+
+	err := os.MkdirAll(filepath.Dir(credPath), 0o700)
+	if err != nil {
+		return err
+	}
+
+	f, err := os.OpenFile(credPath, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0o700)
```
amoeba:
```suggestion
	f, err := os.OpenFile(credPath, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0o600)
```

Do we need user-execute permission? I think user read/write is fine

auth/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (u *Uri) UnmarshalText(text []byte) error {
+	return (*url.URL)(u).UnmarshalBinary(text)
+}
+
+type Credential struct {
+	Type         Type   `toml:"type"`
+	AuthURI      Uri    `toml:"auth_uri"`
+	IndexURI     Uri    `toml:"index_uri"`
```
amoeba: Just to check: To support the user being logged into multiple private registries, would `IndexURI` here be our lookup key when (1) looking for a credential and (2) when running logout with a URI? 

I'm thinking about having super consistent naming everywhere. Right now, this value is just called "URI" in `dbc auth` and "IndexURI" here.

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/columnar-tech/dbc/auth"
+)
+
+type AuthCmd struct {
+	Login  *LoginCmd  `arg:"subcommand" help:"Authenticate with a driver index"`
+	Logout *LogoutCmd `arg:"subcommand" help:"Logout from a driver index"`
+}
+
+type LoginCmd struct {
+	URI      string `arg:"positional" help:"URL of the driver index to authenticate with"`
```
amoeba: I would be really happy if we called this URL instead of URI. Or hostname if we will only ever support top level registries. Going more specific has some advantages for users who are punching in their own info.

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+	ClientID string `arg:"env:OAUTH_CLIENT_ID" help:"OAuth Client ID (can also be set via DBC_OAUTH_CLIENT_ID)"`
+	Audience string `arg:"env:OAUTH_AUDIENCE" help:"OAuth Audience (can also be set via DBC_OAUTH_AUDIENCE)"`
+	ApiKey   string `arg:"--api-key" help:"Authenticate using an API key instead of OAuth"`
```
amoeba: This can be a future feature request after this PR, could we take an API key on stdin just to give users more flexibility?

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+
+type AuthCmd struct {
+	Login  *LoginCmd  `arg:"subcommand" help:"Authenticate with a driver index"`
+	Logout *LogoutCmd `arg:"subcommand" help:"Logout from a driver index"`
+}
+
+type LoginCmd struct {
+	URI      string `arg:"positional" help:"URL of the driver index to authenticate with"`
+	ClientID string `arg:"env:OAUTH_CLIENT_ID" help:"OAuth Client ID (can also be set via DBC_OAUTH_CLIENT_ID)"`
+	Audience string `arg:"env:OAUTH_AUDIENCE" help:"OAuth Audience (can also be set via DBC_OAUTH_AUDIENCE)"`
```
amoeba: Are these temporary, just for testing?

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+	"strings"
+
+	"github.com/charmbracelet/bubbles/spinner"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/cli/oauth/device"
+	"github.com/columnar-tech/dbc/auth"
+)
+
+type AuthCmd struct {
+	Login  *LoginCmd  `arg:"subcommand" help:"Authenticate with a driver index"`
```
amoeba:
```suggestion
	Login  *LoginCmd  `arg:"subcommand" help:"Authenticate with a driver registry"`
```

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+	Logout *LogoutCmd `arg:"subcommand" help:"Logout from a driver index"`
```
amoeba:
```suggestion
	Logout *LogoutCmd `arg:"subcommand" help:"Log out from a driver registry"`
```

cmd/dbc/auth.go
amoeba:
```suggestion
	URI      string `arg:"positional" help:"URL of the driver registry to authenticate with"`
```

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+
+	base, cmd := m.baseModel.Update(msg)
+	m.baseModel = base.(baseModel)
+	return m, cmd
+}
+
+func (m loginModel) View() string { return m.spinner.View() }
+
+type LogoutCmd struct {
+	URI string `arg:"positional" help:"URL of the driver index to logout from"`
```
amoeba:
```suggestion
	URI string `arg:"positional" help:"URL of the driver registry to log out from"`
```

logout -> log out

cmd/dbc/auth.go
amoeba: This should this not be "Log into a driver registry"? The opposite of Authenticate is ... de-authenticate?

cmd/dbc/main.go
```diff
@@ -134,6 +134,7 @@ type cmds struct {
 	Info       *InfoCmd         `arg:"subcommand" help:"Get information about a driver"`
 	Docs       *DocsCmd         `arg:"subcommand" help:"Open driver documentation in a web browser"`
 	Remove     *RemoveCmd       `arg:"subcommand" help:"Remove a driver from the driver list"`
+	Auth       *AuthCmd         `arg:"subcommand" help:"Manage driver index authentication credentials"`
```
amoeba:
```suggestion
	Auth       *AuthCmd         `arg:"subcommand" help:"Manage driver registry credentials"`
```

I think we can drop "authentication" and just say "credentials" since it is implied

.github/workflows/go.yml
```diff
@@ -33,8 +33,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # macos-13 is intel macos while macos-latest is arm64
-        os: [ 'ubuntu-latest', 'windows-latest', 'macos-latest', 'macos-13' ]
+        # macos-15 is intel macos while macos-latest is arm64
+        os: [ 'ubuntu-latest', 'windows-latest', 'macos-latest', 'macos-15' ]
```
amoeba, line 37: How would you feel if I asked you to kick this change out of this PR so we can put it in the releasenotes as a standalone item?

cmd/dbc/auth.go
```diff
@@ -0,0 +1,261 @@
...
+	return tea.Batch(m.spinner.Tick, func() tea.Msg {
+		return u
+	})
+}
+
+func (m loginModel) authConfig() tea.Cmd {
+	return func() tea.Msg {
+		cfg, err := auth.GetOpenIDConfig(m.parsedURI)
+		if err != nil {
+			return fmt.Errorf("failed to get OpenID configuration: %w", err)
```
amoeba, line 117: The error we produce here feels slightly leaky:

```
$ go run ./cmd/dbc auth login
Error: failed to get OpenID configuration: Get "https://dbc-cdn-private.columnar.tech/.well-known/openid-configuration": dial tcp: lookup dbc-cdn-private.columnar.tech: no such host
exit status 1
```

The user doesn't need to know any implementation details of our OAuth flow, not that the URL or the path are secrets and not easily guessable. Thoughts? There's a good chance I'm being needlessly paranoid.

amoeba (commented):
This works really well, thanks for this @zeroshade. I left some comments and suggestions and have some other notes.

1. The user agent on the auth requests is a generic UA. Do we want to use a dbc specific UA string?
2. When I run `dbc auth login` when already logged in, I get a non-zero exit status code. I get a 0 when I log in the first time.

These are some Auth0 setup notes:

1. The Auth0 screen says "Log in to columnar". Can it be Columnar?
2. When I break the oauth flow on purpose, I get an error screen with two issues: Broken logo and lowercase Columnar
    Image
3. My password manager picks up the wrong name for the login, regardless of whether I log in through a Columnar account or a Google account. "Us" which I assume comes from the auth0 subdomain `.us.auth0`. I'm happy to track down why this is for you.

    Image
4. When I finish logging in, I see a page with an incomplete title:

    Image

cmd/dbc/auth.go
lidavidm: Maybe we need to set up logging so that we can use friendly errors and log the details to something that can be shared with us separately?

auth/credentials.go
zeroshade: Go doesn't seem to have a shortcut function for XDG_DATA_HOME in the `os` package like it does for `XDG_CONFIG_HOME` and `XDG_CACHE_HOME`. I'll put something together for it that follows the same pattern as those for us to use.

cmd/dbc/auth.go
amoeba: Yeah that would be really good. I'll create a new issue.

auth/credentials.go
zeroshade: yes, the `IndexURI` is the lookup key

the reason for the differentiation here is that when performing oauth auth flows we have to determine the URI to use for auth by querying from the index (via .well-known paths etc.). To avoid doing that lookup frequently, we cache the index uri AND the auth uri separately. But from the user's perspective, they are only giving us 1 URI, the index uri. I can change the argument in `dbc auth` to be `IndexURI` if that's preferred

cmd/dbc/auth.go
amoeba: https://github.com/columnar-tech/dbc/issues/219

cmd/dbc/auth.go
zeroshade: if we want to allow arbitrary oauth flows to work we need to accept these. Many configurations can set a default audience so I can remove the audience argument, but we'll still need to accept the ClientID if we want to allow arbitrary oauth flows.

amoeba: Okay that makes sense.

auth/credentials.go
amoeba: Thanks, seems fine to keep as-is for now outside of changing "Index" to "Registry" to align with our naming.

cmd/dbc/auth.go
zeroshade: should I change this PR? Or leave it to a follow-up via the issue you filed?

zeroshade:
@amoeba I've addressed your comments and updated the PR.

> The user agent on the auth requests is a generic UA. Do we want to use a dbc specific UA string?

I've updated the user-agent for all the requests to use the dbc user-agent.

> When I run dbc auth login when already logged in, I get a non-zero exit status code. I get a 0 when I log in the first time.

Fixed this!

> The Auth0 screen says "Log in to columnar". Can it be Columnar?

I think I've fixed this, at worst we might need to create a new tenant instead since Auth0 doesn't allow changing the tenant name.

> When I break the oauth flow on purpose, I get an error screen with two issues: Broken logo and lowercase Columnar

How did you break it? This seems like it might be an auth0 error we should file since I've updated the configs to use the logo properly etc.

> My password manager picks up the wrong name for the login, regardless of whether I log in through a Columnar account or a Google account. "Us" which I assume comes from the auth0 subdomain .us.auth0. I'm happy to track down why this is for you.

I've updated our config to use a custom domain via `auth.columnar.tech`, let's see if that fixes that.

Please take another look and review! thanks!

cmd/dbc/auth.go
amoeba: follow up sounds fine. I'm not even 100% sure if we need to quiet this specific error.

amoeba:
Re-tested my above things and the main things are fixed. Thanks @zeroshade. 

I notice that the device confirmation screen now says, "Please confirm this is the code displayed on your dbc Private Registry:". Could that just be "Please confirm this is the code displayed by dbc:"?

And, I think we could tweak the output just a bit. The way we have this set up, the user never needs to know the copy code because we don't do the same kind of thing, say, GitHub or Apple do where they make you enter the copy yourself.

Right now we show:

```sh
$ dbc auth login
Copy code: XXXX-XXXX
To authenticate, visit: https://auth.columnar.tech/activate?user_code=XXXX-XXXX
⠋
```

but it could be even simpler unless you think it's unwise:

```sh
$ dbc auth login
Opening https://auth.columnar.tech/activate?user_code=XXXX-XXXX in your default web browser...
⠋ Waiting for confirmation...

zeroshade:
@amoeba made both changes and updated!

amoeba:
Thanks @zeroshade. I think that we may tweak some parts of this as we continue to discuss but I think we could merge this since it works well.

ianmcook:
I'm trying this now.

ianmcook:
One problem I encountered:

If you run `dbc auth login`, do the OAuth login, then do `dbc auth logout`, then do `dbc auth login` again, it re-uses the same Auth0 browser session and logs you in again without a credential challenge. This is convenient, but the downside is that there's no apparent way no way to change which account you're logging in as. @zeroshade is there a way to solve that?

zeroshade:
You could clear your relevant cookies to get rid of the session. There might be a way to explicitly revoke/end the session though. I'll look into it

zeroshade:
@ianmcook looks like the only way would be to direct them to a URL which explicitly logs them out of the browser since the issue is the browser session cookie. We can probably display a URL to do that but with the auth registry site, they'd also be able to log out of there I believe? I'll poke at this more in the future. For now, I think I'll merge this

ianmcook:
>direct them to a URL which explicitly logs them out of the browser

Cool. What's the URL?

PR #221, opened 2025-12-03
feat: support getting docs urls from driver registry index

### LINKED ISSUE #220: Support docs URLs defined in the driver registry index
The current `dbc docs` implementation relies on a hardcoded mapping between driver paths and URLs but it should really be relying on what's in the driver registry index.  Refactor the current doc docs implementation to look for a docs URL in the index.

The driver registry index doesn't currently have a specific field to hold the URL for a driver's docs so we'll need to implement the above feature in dbc but also,

1. Update the driver registry index YAML to include a field
2. Update the manifests for drivers to include the new field
3. Verify the CDN deployment pulls the new field in

amoeba:
Implements a proper solution for `dbc docs` to get a URL for a driver's docs when the user runs `dbc docs $driver_path`. The previous implementation used a hardcoded set of fallback URLs just to get something working. This feature works gracefully with the current driver registry index structure but can take advantage of a new "docs_url" if one is present.

This PR also updates the fallback URLs map so every driver now works with `dbc docs`.

Closes #220

### DIFF
--- a/cmd/dbc/search_test.go
+++ b/cmd/dbc/search_test.go
@@ -24,12 +24,12 @@ func (suite *SubcommandTestSuite) TestSearchCmd() {
 	m := SearchCmd{}.GetModelCustom(
 		baseModel{getDriverRegistry: getTestDriverRegistry,
 			downloadPkg: downloadTestPkg})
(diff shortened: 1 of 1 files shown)


amoeba:
Rebased and fixed test failures.

PR #223, opened 2025-12-08
feat(auth): automatically fetch Columnar license on login

zeroshade:
When using `dbc auth login` to login to the Columnar private registry, we'll check the token and see if it contains the `trial_start` claim, downloading the license if it exists.

There's some more work that needs to happen here so I'm leaving this marked as WIP, such as:

* Picking a domain for the cloudflare workers
* making the worker domain configurable via env var
* unit tests
* verifying this workflow makes sense

### DIFF
--- a/cmd/dbc/auth.go
+++ b/cmd/dbc/auth.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -210,6 +211,16 @@ func (m loginModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				}
 				return tea.Quit()
 			})
+	case error:
+		switch {
+		case errors.Is(msg, auth.ErrTrialExpired) ||
+			errors.Is(msg, auth.ErrNoTrialLicense):
+			// ignore these errors during auth login
+			// the user can still login but won't be able to download trial licenses
+			return m, tea.Quit
+		default:
+			// for other errors, let the baseModel update handle it.
+		}
 	}
 
 	base, cmd := m.baseModel.Update(msg)
--- a/cmd/dbc/main.go
+++ b/cmd/dbc/main.go
@@ -120,6 +120,7 @@ func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, tea.Quit
 		}
 	case error:
+		m.status = 1
 		var cmd tea.Cmd
 		switch {
 		case errors.Is(msg, auth.ErrTrialExpired):
@@ -127,7 +128,6 @@ func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		case errors.Is(msg, auth.ErrNoTrialLicense):
 			cmd = tea.Println(errStyle.Render("Could not download license, trial not started"))
 		default:
-			m.status = 1
 			cmd = tea.Println("Error: ", msg.Error())
 		}
 		return m, tea.Sequence(cmd, tea.Quit)


ianmcook:
Nice! Let's also please make it so that if the user runs `dbc install` or `dbc sync` to install a private driver, dbc checks to see whether:
1. the user is authenticated
2. the user has an active trial license
3. the user has _not_ yet downloaded `columnar.lic` and put it in the required location.

If all three of those things are true, dbc should download `columnar.lic` and put it in the required location for the user.

zeroshade:
@ianmcook updated per your suggestion

zeroshade:
Added @lidavidm so he can see where dbc is putting the license

ianmcook:
> @ianmcook updated per your suggestion

Nice! I tested it successfully.

lidavidm (commented):
If you log in and your trial is expired, you don't get any feedback:

```
$ dlv debug --headless ./cmd/dbc -- auth login
API server listening at: 127.0.0.1:39367
Opening https://auth.columnar.tech/activate?user_code=QVLR-GDCV in your default web browser...
Gtk-Message: 13:10:44.101: Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
Authentication successful!
```

In `dlv`, I can see that I got a 403:

```
> github.com/columnar-tech/dbc/auth.FetchColumnarLicense() ./auth/credentials.go:318 (PC: 0x978c8a)
   313:		if err != nil {
   314:			return err
   315:		}
   316:		defer resp.Body.Close()
   317:	
=> 318:		if resp.StatusCode != http.StatusOK {
   319:			return fmt.Errorf("failed to fetch license: %s", resp.Status)
   320:		}
   321:	
   322:		licenseFile, err := os.OpenFile(licensePath, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0o600)
   323:		if err != nil {
(dlv) p resp.StatusCode
403
```

But this is not reported in the UI.

amoeba:
I'm not sure if this is the same error as @lidavidm got but here's what I did:

1. Ran go run ./cmd/dbc auth login
2. In my browser, saw the user code, clicked Confirm
3. Logged in to Columnar with Google
4. Got to my account details page (https://cloud.columnar.tech/account)

Back in my terminal, the CLI is stuck on 

> Authentication successful!
> ⠙ Waiting for confirmation...

I note that my trial is expired. Also I notice that the text says 

> Expires: Thu, 25 Dec 2025 23:16:09 GMT

but it should probably tell me that my license is _expired_.

cmd/dbc/auth.go
```diff
@@ -190,13 +194,22 @@ func (m loginModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+	case authSuccessMsg:
+		return m, tea.Sequence(tea.Println("Authentication successful!"),
+			func() tea.Msg {
+				if auth.IsColumnarPrivateRegistry((*url.URL)(&msg.cred.RegistryURL)) {
+					return auth.FetchColumnarLicense(&msg.cred)
+				}
+				return nil
+			})
+	case nil:
+		return m, tea.Quit
```
amoeba: This looked fine to me but my auth command hung indefinitely. I think bubbletea doesn't like nil being used a signal.
```suggestion
				return tea.Quit()
			})
```

zeroshade:
@amoeba can you try again? It looks good to me (and I'll update the website to be more explicit when the trial is expired)

amoeba:
Thanks @zeroshade. I can't reproduce anymore as I don't have an expired license but `dbc auth login` works fine now.

auth/credentials.go
```diff
@@ -265,3 +267,75 @@ func UpdateCreds() error {
...
+		}
+		authToken = "Bearer " + cred.GetAuthToken()
+	}
+
+	req, err := http.NewRequest(http.MethodGet, licenseURI, nil)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("authorization", "Bearer "+authToken)
```
amoeba: It looks like this would lead to having two "Bearer" strings in the output. What do you think?

auth/credentials.go
```diff
@@ -265,3 +267,75 @@ func UpdateCreds() error {
...
+		p := jwt.NewParser()
+		tk, err := p.Parse(cred.GetAuthToken(), nil)
+		if err != nil && !errors.Is(err, jwt.ErrTokenUnverifiable) {
+			return fmt.Errorf("failed to parse oauth token: %w", err)
+		}
+
+		_, ok := tk.Claims.(jwt.MapClaims)["urn:columnar:trial_start"]
+		if !ok {
+			return ErrNoTrialLicense
+		}
```
amoeba, line 307: In theory this could panic as written unless you do this right?

```suggestion
		claims, ok := tk.Claims.(jwt.MapClaims)
		if !ok {
			return ErrNoTrialLicense
		}
		_, ok = claims["urn:columnar:trial_start"]
		if !ok {
			return ErrNoTrialLicense
		}
```

auth/credentials.go
```diff
@@ -265,3 +267,75 @@ func UpdateCreds() error {
...
+			return fmt.Errorf("failed to fetch license: %s", resp.Status)
+		}
+	}
+
+	licenseFile, err := os.OpenFile(licensePath, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0o600)
+	if err != nil {
+		return err
+	}
+	defer licenseFile.Close()
+	_, err = licenseFile.ReadFrom(resp.Body)
```
amoeba: What happens to the user if this write files? Either they end up with an empty file or a partial file. Does dbc handle that well (can the user self-rescue) or should we maybe handle `err` in place and remove `licenseFile` if `err != nil`?

auth/credentials.go
```diff
@@ -265,3 +267,75 @@ func UpdateCreds() error {
...
+		if err != nil && !errors.Is(err, jwt.ErrTokenUnverifiable) {
+			return fmt.Errorf("failed to parse oauth token: %w", err)
+		}
+
+		_, ok := tk.Claims.(jwt.MapClaims)["urn:columnar:trial_start"]
+		if !ok {
+			return ErrNoTrialLicense
+		}
+		authToken = "Bearer " + cred.GetAuthToken()
+	}
```
amoeba, line 311: Add a default case that returns error?

cmd/dbc/main.go
```diff
@@ -118,8 +120,17 @@ func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, tea.Quit
 		}
 	case error:
-		m.status = 1
-		return m, tea.Sequence(tea.Println("Error: ", msg.Error()), tea.Quit)
+		var cmd tea.Cmd
+		switch {
+		case errors.Is(msg, auth.ErrTrialExpired):
+			cmd = tea.Println(errStyle.Render("Could not download license, trial has expired"))
```
amoeba, line 127: set non-zero status? I'm not totally sure if we should or not.

cmd/dbc/main.go
```diff
@@ -118,8 +120,17 @@ func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		case errors.Is(msg, auth.ErrNoTrialLicense):
+			cmd = tea.Println(errStyle.Render("Could not download license, trial not started"))
```
amoeba, line 129: Set non-zero status?

auth/credentials.go
zeroshade: wow, good catch! how the hell did I not come across this?

auth/credentials.go
zeroshade: In theory, yes. But we're pretty much guaranteed that if it successfully parses then it *will* be a `jwt.MapClaims` type because we aren't providing a custom type for it to use. So we don't need to do the additional check

auth/credentials.go
zeroshade: that's a good point, i'll have it check the err in place and remove the file if err != nil

cmd/dbc/main.go
zeroshade: My personal opinion is that we don't need a non-zero status for the trial handling because I don't want dbc to always fail if they login but haven't started the trial yet etc.

zeroshade: okay, after discussing with @ianmcook we decided that this will give a non-zero status, but it won't error/give a non-zero status during `dbc auth login`. It'll only error during install/sync

I've updated this accordingly

auth/credentials.go
amoeba: I like this "pretty much guaranteed" level of assurance :)

Okay as is.

auth/credentials.go
amoeba: Thanks. Sorry for my typo above s/files/fails.

amoeba:
I ran into an issue while testing driver list support for this,

```
$ go run ./cmd/dbc info red-o
Driver: red-o
Version: 0.3.0
Title: Columnar ADBC Driver red-o
License: LicenseRef-PBL
Description: An ADBC driver for Oracle Database developed by Columnar
Available Packages:
   - linux_amd64
   - linux_arm64
   - macos_arm64
   - windows_amd64
   
~/src/columnar-tech/dbc auth-fetch-license
$ go run ./cmd/dbc init

~/src/columnar-tech/dbc auth-fetch-license
$ go run ./cmd/dbc add red-o
added red-o to driver list
use `dbc sync` to install the drivers in the list

~/src/columnar-tech/dbc auth-fetch-license
$ go run ./cmd/dbc sync
Error: failed to download driver: failed to download driver https://dbc-cdn-private.columnar.tech/red-o/v0.3.0/red-o_macos_arm64_v0.3.0.tar.gz: 401 Unauthorized
exit status 1
```

This was `go run`ning from `7a4e818e9e20269d133df301dbfa9cf63d2c1de7`.

amoeba:
I ran dbc auth login and it worked after I did that. I can't remember how I got the token that 401's but, looking at it decoded, the biggest difference is that,

- The token that 401s had `"permissions":[]`
- The token that worked fine had `"permissions":["access:index","access:red-o"]`

Otherwise things looked structurally similar.

zeroshade:
*sigh* yea, looks like I might have to do part of what I was discussing with Ian. I was trying to reduce calls and latency but I guess there's no way around it. I'll do some stuff tomorrow to make this more robust and handle things better instead of relying on the (potentially stale) login token

zeroshade:
@amoeba Okay, I've updated the lambda func to fallback and make a call to heimdall.columnar.tech for the latest token when the token came from Auth0 and doesn't contain the appropriate permissions. This will cover the scenario you had where you had to re-login.

amoeba:
Nice! Let me test that out in a sec and I'll report back.

amoeba:
I wasn't able to get this to work how I expected. I deleted my account in auth0, signed up with a fresh account, logged in but did not start my trial, ran `dbc auth login`, started my trial, and ran `dbc install oracle`. It hung for a good 5-10 seconds and then error'd with,

```
$ go run ./cmd/dbc install oracle
Error: failed to download driver https://dbc-cdn-private.columnar.tech/oracle/v0.4.0/oracle_macos_arm64_v0.4.0.tar.gz: 401 Unauthorized
exit status 1
```

amoeba:
I had mitmproxy running dbc made two requests (amongst others) to install the driver. The first 401 had a JWT with `"permissions":[]` as expected. The second 401 had `"permissions":["access:index","access:red-o"]}`

amoeba:
Maybe something else is going on. I went back to the preview Cloud site and the UI didn't show I had a trial. I then tried to restart my trial and it kept failing after I hit Accept with a 400 "Trial already started" in the console. I'll file a bug since the registry site needs to handle that.

PR #224, opened 2025-12-09
docs: add cheatsheet

esadek:
Add the cheat sheet to the documentation.

image

### DIFF
--- a/docs/getting_started/cheatsheet.md
+++ b/docs/getting_started/cheatsheet.md
@@ -20,6 +20,6 @@ The dbc cheatsheet covers commands for installing dbc, driver discovery, driver 
 
 <a href="/dbc/assets/cheatsheet.pdf" class="md-button md-button--primary">Download</a>
 
-<img src="/dbc/assets/cheatsheet.png" />
+[![dbc cheatsheet](/dbc/assets/cheatsheet.png)](/dbc/assets/cheatsheet.pdf)
 
 This work is licensed under [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/).


amoeba (commented):
Thanks for this. What are the downsides with embedding the PDF like this? I wonder if we should just use a sized-down screenshot and link to it. It'd also be nice to explicitly call out the license and give a clear download link.

esadek:
I embedded the PDF because it displays the cheat sheet and includes a download button (via the browser).
After some thought, I agree that having an image and download link would be cleaner. Most of the functionality of the built-in PDF viewer is not at all useful (in Firefox at least). Will add mention of the license as well.

docs/getting_started/cheatsheet.md
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Cheatsheet
+
+The dbc cheatsheet covers commands for installing dbc, driver discovery, driver management, and driver lists.
+
+<a href="/dbc/assets/cheatsheet.pdf" class="md-button md-button--primary">Download</a>
```
amoeba, line 21:
```suggestion
[Download PDF](/dbc/assets/cheatsheet.pdf){ .md-button .md-button--primary }
```

nit: slightly preferable to use markdown everywhere we can

docs/getting_started/cheatsheet.md
```diff
@@ -0,0 +1,25 @@
...
+
+<img src="/dbc/assets/cheatsheet.svg" />
```
amoeba: This SVG ended up kinda big (142KiB)  and it doesn't look like it was saved with good default colors so it just looks like this locally:

Image

Can we just save it as a PNG?

amoeba: can we make this a link so the user can click it to get the pdf?

amoeba (changes requested):
Thanks. The experience across browsers with embedding PDFs is not great so I think the image is better. Safari is a good example.

I pushed up one change (6dbe1478b856b353a831eef2e6343fdb661e55cd) and had just a few more comments. I'll be sure to review promptly once you go through them. Thanks for humoring me :)

docs/getting_started/cheatsheet.md
esadek: This doesn't seem to be supported

image

amoeba: Too bad. Thanks for testing.

docs/getting_started/cheatsheet.md
esadek: Strange, it renders fine on my machine.
Changed it to a PNG.

esadek: Good idea. Added a link to the PDF and changed it to Markdown.

PR #225, opened 2025-12-10
feat(registry): multi-registry support

zeroshade:
Enable `GetDriversList` to use multiple registries as source for the driver list, defaulting to both the public and dbc-cdn-private registries. Will ignore a registry if it receives a 401 Unauthorized response from the index.yaml request.

This does not yet provide a way to configure the available registries, I'll add that in a follow-up

### DIFF
--- a/drivers.go
+++ b/drivers.go
@@ -39,7 +39,7 @@ import (
 	machineid "github.com/zeroshade/machine-id"
 )
 
-type Index struct {
+type Registry struct {
 	Name    string
 	Drivers []Driver
 	BaseURL *url.URL
@@ -54,7 +54,7 @@ func mustParseURL(u string) *url.URL {
 }
 
 var (
-	registries = []Index{
+	registries = []Registry{
 		{BaseURL: mustParseURL("https://dbc-cdn.columnar.tech")},
 		{BaseURL: mustParseURL("https://dbc-cdn-private.columnar.tech")},
 	}
@@ -82,6 +82,12 @@ func init() {
 	info, ok := debug.ReadBuildInfo()
 	if ok {
 		Version = info.Main.Version
+	}
+
+	if val := os.Getenv("DBC_BASE_URL"); val != "" {
+		registries = []Registry{
+			{BaseURL: mustParseURL(val)},
+		}
 	}
 
 	userAgent := fmt.Sprintf("dbc-cli/%s (%s; %s)",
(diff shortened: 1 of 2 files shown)


auth/credentials_test.go
lidavidm, line 1: It's more normal to have the "Bearer" prefix, does something not support that?

zeroshade: I'm using the lack of Bearer to denote the difference between the apikey and the JWT. 

I did a bunch of research and it looks like ApiKeys will typically drop the bearer or use a prefix like "token", or drop the "Authorization" header entirely and use a custom header like "X-Api-Key" instead. I'm up for changing this if we think it's worth it. I was just trying to make life easier to denote between the jwt token and the api key

lidavidm: Ah ok sounds good, then

drivers.go
```diff
@@ -39,10 +39,25 @@ import (
 	machineid "github.com/zeroshade/machine-id"
 )
 
-const defaultURL = "https://dbc-cdn.columnar.tech"
+type Index struct {
```
amoeba: Do we want to use the word "Registry" here to align with the terminology we use in the docs?

drivers.go
```diff
@@ -39,10 +39,25 @@ import (
...
+	uri, err := url.Parse(u)
+	if err != nil {
+		panic(fmt.Sprintf("failed to parse URL %s: %v", u, err))
+	}
+	return uri
+}
 
 var (
-	baseURL = defaultURL
+	registries = []Index{
```
amoeba: so this would be

```suggestion
	registries = []Registry{
```

drivers.go
```diff
@@ -69,10 +84,6 @@ func init() {
 		Version = info.Main.Version
 	}
 
-	if val := os.Getenv("DBC_BASE_URL"); val != "" {
```
amoeba, line 72: It would be nice to keep an env var override for testing.

auth/credentials_test.go
amoeba: I think both are technically Bearer tokens in the sense the bearer means "if I have this key, I have access". And, if I'm reading this right, a schema (Bearer being an auth-scheme) is required? https://www.rfc-editor.org/rfc/rfc9110.html#section-11.4. The more common practice I've seen where you need to know what type of token a token is is to prefix them (like GitHub does).

ianmcook:
This seems to work great. Thanks @zeroshade!

drivers.go
zeroshade: added the env var back

drivers.go
zeroshade: done

drivers.go
zeroshade: sure, i renamed the type to `Registry`

amoeba: Thanks

PR #231, opened 2025-12-12
fix: fix incorrect credentials path, create common config logic

### LINKED ISSUE #230: Rationalize locations where dbc configs are stored
Currently on macOS:
- `uid.uuid` is created in `~/Library/Application Support/Columnar/dbc`
- The `credentials` subdirectory (which contains `columnar.lic` and `credentials.toml`) is created in into `~/Library/dbc`

Shouldn't we have these be in the same place?

amoeba:
Closes #230

### DIFF
--- a/internal/dirs.go
+++ b/internal/dirs.go
@@ -1,0 +1,25 @@
+package internal
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// Get a platform-specific config dir for reading and writing dbc config files
+// and credentails
+func GetDbcConfigDir() (string, error) {
+	userdir, err := os.UserConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get dbc configuration directory: %v", err)
+	}
+
+	orgDirName := "columnar"
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		orgDirName = "Columnar"
+	}
+	dbcDirName := "dbc"
+
+	return filepath.Join(userdir, orgDirName, dbcDirName), nil
+}
--- a/auth/credentials.go
+++ b/auth/credentials.go
@@ -22,10 +22,10 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"runtime"
 	"slices"
 	"sync"
 
+	"github.com/columnar-tech/dbc/internal"
 	"github.com/pelletier/go-toml/v2"
 )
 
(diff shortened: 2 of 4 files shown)


internal/dirs.go
```diff
@@ -0,0 +1,25 @@
...
+	userdir, err := os.UserConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get dbc configuration directory: %v", err)
+	}
+
+	orgDirName := "columnar"
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		orgDirName = "Columnar"
+	}
+	dbcDirName := "dbc"
```
zeroshade, line 22: but you were the one who said I should use `XDG_DATA_HOME` if it exists.... lol

amoeba: Oh right. 🤦  This still fixes the original issue that was reported. Let me think about it and I'll tweak this so it uses XDG_DATA_DIR on Linuxes.

amoeba: Which would make this:

- macOS + Windows: os.UserConfigDir()
- Linux: check XDG_DATA_DIR, fallback to os.UserConfigDir()

zeroshade: it should be `XDG_DATA_HOME`, there is no `XDG_DATA_DIR` there is `XDG_DATA_DIRS` which is a list of paths

zeroshade: and we should still check `XDG_DATA_HOME` on macOS too. According to Apple's docs, the equivalent for `XDG_DATA_HOME` on macOS is `~/Library/`

amoeba: Oh, that last bit I didn't know.

amoeba: Maybe my idea was bad, I'll think about it more when I'm back.

zeroshade: yea, that's why it was using the logic I laid out. which was separate from the XDG_CONFIG_HOME (os.UserConfigDir) heh

zeroshade:
Did we come to a consensus on this?

amoeba:
> Did we come to a consensus on this?

I wasn't able to find anything solid to go off of. The only thing that recommends a dir like `~/Library/dbc` is a random FAQ on Java: https://developer.apple.com/library/archive/qa/qa1170/_index.html. But it makes a point that aligns with how XDG does things: it's good practice to separate config from data. Is this the ref you are referring to when you say,

> According to Apple's docs, the equivalent for XDG_DATA_HOME on macOS is ~/Library/...

I think I'm okay with closing this PR unmerged and keeping the current behavior.

amoeba:
I just created a PR with regression tests for the current credentials path behavior: https://github.com/columnar-tech/dbc/pull/247.

ianmcook:
After more discussion in #230, I think the decision is:

UUIDs, other config, etc. go under:

* macOS: `$HOME/Library/Application Support`
* Linux: `$XDG_CONFIG_HOME` or `$HOME/.config`
* Windows: `%APPDATA%` (i.e. `C:\Users\<Username>\AppData\Roaming`)

Credentials go under:

* macOS: `$HOME/Library/Application Support` (same as above)
* Linux: `$XDG_DATA_HOME` or `$HOME/.local/share`
* Windows: `%LOCALAPPDATA%` (i.e. `C:\Users\<Username>\AppData\Local`)

PR #238, opened 2025-12-17
chore: support testing user and system config levels in CI

### LINKED ISSUE #210: Enable testing user and system level configs in ci
Our unit test suite mostly exercises the env config level, so as not to leave the user's system in a weird state. This is good.

However, we still need a way to do unit and probably snapshot tests against user and system levels. It would help us find bugs like https://github.com/columnar-tech/dbc/issues/203. We can do this in CI.

amoeba:
Adds a mechanism for running tests for the user and system config levels and adds a basic installation test for both config levels. The goal here is not to completely replicate the existing tests but to give us a way to test special cases of interest.

The approach here creates a new test suite class and standalone files instead of mixing them into other test suites and they are intentionally hard to run because they are potentially dangerous. And I wasn't able to find a good mechanism to do something more like Python or Rust where it's easier to mix tests together. 

Tests are in `integration_user_test.go` and `integration_system_test.go`. I split these up because, in CI, I want to test the user level tests without sudo and the system level tests _with_ sudo. There are other ways to work around this but I feel better just testing what the users would run directly.

You have to run the tests like, 

```console
# user
$ DBC_RUN_INTEGRATION_TESTS=1 go test -tags test_integration,test_system
#system
$ DBC_RUN_INTEGRATION_TESTS=1 go test -tags test_integration,test_user
```

This PR enables them in CI for all commits.

Closes #210

### DIFF
--- a/cmd/dbc/integation_user_test.go
+++ b/cmd/dbc/integation_user_test.go
@@ -29,7 +29,7 @@ func (suite *IntegrationTestSuite) TestInstallUser() {
 
 	loc := config.GetLocation(config.ConfigUser)
 	suite.Equal("\nInstalled test-driver-1 1.1.0 to "+loc+"\n", out)
-	suite.driverIsInstalled(config.ConfigUser, "test-driver-1", "1.1.0")
+	suite.driverIsInstalled(config.ConfigUser, "test-driver-1")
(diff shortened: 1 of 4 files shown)


amoeba:
Closing in favor of https://github.com/columnar-tech/dbc/pull/239.

PR #239, opened 2025-12-19
chore: parameterize SubCommandTest with configLevel

### LINKED ISSUE #210: Enable testing user and system level configs in ci

### LINKED ISSUE #215: Driver manifest symbolic link not deleted by uninstall command
After uninstalling a driver with `dbc uninstall`, the symbolic link to the driver manifest is still present.

```console
$ pwd
/Users/emilsadek/Library/Application Support/ADBC

$ ls

$ dbc install duckdb
[✓] searching
[✓] downloading
[✓] installing
[✓] verifying signature

Installed duckdb 1.4.2 to /Users/emilsadek/Library/Application Support/ADBC/Drivers

$ ls
Drivers         duckdb.toml

$ dbc uninstall duckdb
Driver `duckdb` uninstalled successfully!

$ ls
Drivers         duckdb.toml
```

OS: macOS 15.6.1 24G90 arm64
dbc: v0.1.0

amoeba:
Extends the SubCommandTest test suite to be parameterized by configLevel so the subcommand tests can be run with `--level` set to user or system by running the test suite with `DBC_TEST_LEVEL_USER=1` or `DBC_TEST_LEVEL_SYSTEM=1` respectively.

This is not enabled by default on developer machines but is enabled in CI. To opt in this behavior, tests need to be written to use `suite.configLevel` to control their config level and tests can opt out either by not doing this or by doing this but adding appropriate Skips.

To demo what we can do now, this PR also includes a fix for #215. We could have fixed the issue by testing in a tempdir but with the change in this PR we can actually test the bug that was reported. Doing this also uncovered some edge cases in the symlink behavior so this was time well spent.

Also fixes a minor undiscovered bug in removing manifest-only drivers: https://github.com/columnar-tech/dbc/pull/239/commits/e69744f3c2ead08099dfb15f6b91a5296ad4fb98.

Supersedes #238
Closes #210 
Closes #215

### DIFF
--- a/config/dirs_windows.go
+++ b/config/dirs_windows.go
@@ -58,7 +58,7 @@ func (c ConfigLevel) rootKeyString() string {
 	}
 }
 
-func (c ConfigLevel) configLocation() string {
+func (c ConfigLevel) ConfigLocation() string {
 	var prefix string
 	switch c {
 	case ConfigSystem:


cmd/dbc/subcommand_test.go
```diff
@@ -83,24 +84,26 @@ type SubcommandTestSuite struct {
...
 }
 
 func (suite *SubcommandTestSuite) SetupTest() {
 	suite.tempdir = suite.T().TempDir()
 	suite.Require().NoError(os.Setenv("ADBC_DRIVER_PATH", suite.tempdir))
 }
 
-func (suite *SubcommandTestSuite) TearDownTest() {
-	suite.Require().NoError(os.Unsetenv("ADBC_DRIVER_PATH"))
-}
```
zeroshade, line 102: why remove this? If we're not going to remove this env var after each test, then we should move the `os.Setenv` to `SetupSuite` and add `Unsetenv` to `TearDownSuite`

cmd/dbc/subcommand_windows_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/columnar-tech/dbc/config"
+	"golang.org/x/sys/windows/registry"
+)
+
+func (suite *SubcommandTestSuite) TearDownTest() {
+	suite.Require().NoError(os.Unsetenv("ADBC_DRIVER_PATH"))
+
+	// Clean up the registry and filesystem after each test
+	_, user := os.LookupEnv("DBC_TEST_LEVEL_USER")
+	_, system := os.LookupEnv("DBC_TEST_LEVEL_SYSTEM")
```
zeroshade, line 31: what happens if both are set?

config/config.go
```diff
@@ -97,6 +97,10 @@ func (c ConfigLevel) String() string {
 	}
 }
 
+func GetLocation(lvl ConfigLevel) string {
+	return lvl.configLocation()
+}
```
zeroshade: why not just export `lvl.configLocation` instead?

config/dirs_windows.go
```diff
@@ -345,6 +345,9 @@ func UninstallDriver(cfg Config, info DriverInfo) error {
 		if err := os.Remove(manifest); err != nil {
 			return fmt.Errorf("error removing manifest %s: %w", manifest, err)
 		}
+
+		// TODO: Remove this when the driver managers are fixed (>=1.8.1).
```
zeroshade, line 349: is there an issue on arrow-adbc to track this?

cmd/dbc/subcommand_test.go
```diff
@@ -83,24 +84,26 @@ type SubcommandTestSuite struct {
...
 	suite.fallbackDriverDocsUrl = fallbackDriverDocsUrl
+
+	if suite.configLevel == config.ConfigUnknown {
+		suite.configLevel = config.ConfigEnv
+	}
 }
 
 func (suite *SubcommandTestSuite) SetupTest() {
 	suite.tempdir = suite.T().TempDir()
 	suite.Require().NoError(os.Setenv("ADBC_DRIVER_PATH", suite.tempdir))
```
zeroshade: actually, we could save ourselves some trouble and just use [`suite.T().Setenv`](https://pkg.go.dev/testing#T.Setenv), let's do that

.github/workflows/go.yml
```diff
@@ -59,3 +59,20 @@ jobs:
         # the file with `test_registry` also requires windows, so we can safely add this tag
         # without it running on non-windows OSes
         run: go test -v -tags test_registry ./...
+
+      # User and System level tests
+      - name: Run Tests (User)
+        run: go test -v ./...
+        env:
+          DBC_TEST_LEVEL_USER: 1
```
zeroshade, line 67: do we need one of these with `DBC_TEST_LEVEL_ENV: 1`?

I also really don't like this personally, but can't think of a better option other than maybe using `TestMain` and adding explicit command line args that we look for instead? meh, I can live with this for now.

cmd/dbc/subcommand_test.go
amoeba: This got split out into platform specific TearDownTest methods (subcommand_unix_test.go, subcommand_windows_test.go)

config/config.go
amoeba: The last time you and I were looking at doing this, I think you mentioned not wanting to export lvl.configLocation so I exported a wrapper here instead. I'll remove the wrapper since I think it's fine just to export lvl.configLocation.

cmd/dbc/subcommand_windows_test.go
amoeba: If both are set, cleanup for both should run. Does that sound right?

config/dirs_windows.go
amoeba: This was fixed in https://github.com/apache/arrow-adbc/commit/526dd933c2a7fca33486133e05e180a7b1a9d1e2. We can file an issue here on dbc to do something about it.

amoeba: https://github.com/columnar-tech/dbc/issues/241

cmd/dbc/subcommand_test.go
amoeba: Good call. I didn't know about that.

I realized we can make the change more comprehensively too so I did that in 643f102ef1483daed5fdcb78f09d621765ac4cff.

config/config.go
amoeba: Done in b7a32891c7c43b89b06edf4b6ce76564cebae75e. `level.ConfigLocation` is now exported.

.github/workflows/go.yml
amoeba: We don't, just running the tests normally (which we do above) tests DBC_TEST_LEVEL_ENV.

This was my third attempt at this and I dislike this approach the least. I can live with it too :)

amoeba:
Thanks for the review @zeroshade. Ready for a re-review.

PR #246, opened 2026-01-10
feat(info, search, install, uninstall): add json output option

zeroshade:
Add `--json` options for output of info and search subcommands

### DIFF
--- a/cmd/dbc/completions/dbc.bash
+++ b/cmd/dbc/completions/dbc.bash
@@ -62,7 +62,7 @@ _dbc_install_completions() {
         --level|-l)
             COMPREPLY=($(compgen -W "user system" -- "$cur"))
             return 0
-            ;;        
+            ;;
     esac
 
     if [[ "$cur" == -* ]]; then


eitsupi (commented):
Seems nice!
Shell completion needs to be updated.

cmd/dbc/info.go
```diff
@@ -23,12 +24,14 @@ import (
 
 type InfoCmd struct {
 	Driver string `arg:"positional,required" help:"Driver to get info about"`
+	Json   bool   `help:"Output JSON instead of plaintext"`
```
amoeba:
```suggestion
	Json   bool   `help:"Print output as JSON instead of plaintext"`
```

Just to keep style consistent with other flags.

cmd/dbc/search.go
```diff
@@ -36,21 +37,24 @@ var (
 
 type SearchCmd struct {
 	Verbose bool           `arg:"-v" help:"Enable verbose output"`
+	Json    bool           `help:"Output JSON instead of plaintext"`
```
amoeba:
```suggestion
	Json    bool           `help:"Print output as JSON instead of plaintext"`
```

amoeba (commented):
Works great, nice one.

Two suggestions for style and we should also update the shell completions and docs.

Here's a patch for the docs,

```patch
diff --git a/docs/reference/cli.md b/docs/reference/cli.md
index 1137183..f8bb8b8 100644
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -66,6 +66,10 @@ $ dbc search [FILTER]
 
 <h3>Options</h3>
 
+`--json`
+
+:   Print output as JSON instead of plaintext
+
 `--verbose`, `-v`
 
 :   Enable verbose output
@@ -255,6 +259,10 @@ $ dbc info <DRIVER>
 
 <h3>Options</h3>
 
+`--json`
+
+:   Print output as JSON instead of plaintext
+
 `--quiet`, `-q`
 
 :   Suppress all output

```

amoeba (approved):
Is this ready for another review? I think this looks good now so +1.

cmd/dbc/completions/dbc.bash
```diff
@@ -62,11 +62,11 @@ _dbc_install_completions() {
         --level|-l)
             COMPREPLY=($(compgen -W "user system" -- "$cur"))
             return 0
-            ;;
+            ;;
```
eitsupi: Maybe unnecessary spaces?

PR #248, opened 2026-01-12
fix(search): improve output for non-verbose

zeroshade:
supercedes #222 

Improves the output of `dbc search` when not using verbose mode:

image

### DIFF
--- a/cmd/dbc/search.go
+++ b/cmd/dbc/search.go
@@ -116,7 +116,8 @@ func viewDrivers(d []dbc.Driver, verbose bool) string {
 	installedStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("241"))
 
 	l := list.New()
-	t := table.New().Border(lipgloss.HiddenBorder())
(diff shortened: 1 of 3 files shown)


cmd/dbc/search.go
```diff
@@ -115,6 +116,7 @@ func viewDrivers(d []dbc.Driver, verbose bool) string {
 	installedStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("241"))
 
 	l := list.New()
+	t := table.New().Border(lipgloss.HiddenBorder())
```
amoeba: It seems like this adds a space on all sides of the output and I don't think I like it. Is there a way to get the table behavior without (1) the blank lines above and below and (2) the preceding whitespace on each line?

zeroshade: See the commit I just pushed :) found the same issue when trying to fix the tests and fixed it

amoeba: Nice, that did it.

amoeba (approved):
LGTM, thank you!

PR #249, opened 2026-01-13
chore: add pre-commit hooks

esadek:
Add pre-commit hooks to the pre-commit config. Also, run pre-commit in CI.

These hooks all run successfully:

```console
$ pre-commit run --all-files                 
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
detect private key.......................................................Passed
mixed line ending........................................................Passed
trim trailing whitespace.................................................Passed
typos....................................................................Passed
```


zeroshade:
We should definitely add this to the CI workflows

esadek:
Any preference for how we do this in CI?

There is an official [pre-commit/action](https://github.com/pre-commit/action), although it is in maintenance-only mode. They recommend using their [pre-commit.ci](https://pre-commit.ci/) service which is free for open source repos.

We could install Python and pre-commit, then run it for all files. `integration.yml` already installs Python, but `dev_pr.yml` does not. Not sure which workflow is best. Another option would be to use Pixi.

zeroshade:
Pixi makes sense, you could also look at the arrow-adbc repo and see how @lidavidm set it up there, and just use the same method :)

amoeba:
I'd put a vote in for doing it just how it's done here: https://github.com/adbc-drivers/dev/blob/main/.github/workflows/dev.yaml. It helps to reduce context switching cost to have it be same or similar. Can that work?

lidavidm:
You should even be able to directly use the generated workflows for this (it shouldn't generate anything outside of dev*.pr if you don't enable any languages)

zeroshade:
@esadek anything else to be done here?

esadek:
@zeroshade Should be ready to merge unless anyone wants changes.

zeroshade:
If no one has any issues, feel free to merge @amoeba

amoeba:
Merged, thanks @esadek.

PR #251, opened 2026-01-19
feat(auth): add the "--purge" option to logout

zeroshade:
Add a `--purge` option to `dbc auth logout` which will delete the entire credentials dir (including the `columnar.lic` and `credentials.toml` files).

Also, it appears that the auth subcommands were not in the autocompletion scripts so this also adds the appropriate completions.

### DIFF
--- a/auth/credentials.go
+++ b/auth/credentials.go
@@ -269,7 +269,21 @@ func UpdateCreds() error {
 }
 
 func PurgeCredentials() error {
-	return os.RemoveAll(filepath.Dir(credPath))
+	var fileList = []string{
+		"credentials.toml",
+		"columnar.lic",
+	}
+
+	prefix := filepath.Dir(credPath)
+
+	for _, file := range fileList {
+		fullPath := filepath.Join(prefix, file)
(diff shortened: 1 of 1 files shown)


auth/credentials.go
```diff
@@ -268,6 +268,10 @@ func UpdateCreds() error {
 	})
 }
 
+func PurgeCredentials() error {
+	return os.RemoveAll(filepath.Dir(credPath))
```
amoeba: I think it's okay to not get too paranoid about this line of code since, while `credPath` can be overridden by the user with env var, we always append multiple subdirs to it so even a path traversal attack would be less successful. Thoughts?

amoeba (approved):
Looks good. One comment.

auth/credentials.go
ianmcook: +1, it'd be safer to delete specific files by name

zeroshade: How's the updated version?

ianmcook:
A simple test would be nice. I'm worried that we'll add more files to this location later then forget to have the `--purge` implementation remove them. @zeroshade can you add a simple test that would catch this?

PR #255, opened 2026-01-20
feat(install): give hint when unauthorized

### LINKED ISSUE #253: Give users a hint about what to do when installing a driver gives a 401 Unauthorized error
Currently if you're not logged in and you try to install a driver that requires you to be logged in, you see:

```
Error: failed to download driver <uri>: 401 Unauthorized
```

But this doesn't give the user any clue what to do. Could we expand that error message to be something like this:

```
Error: failed to download driver <uri>: 401 Unauthorized
Did you run `dbc auth login`?
```

?

zeroshade:
closes #253


ianmcook:
I tested this and it looks good to me! @amoeba any suggestions here before Matt merges it?

PR #256, opened 2026-01-20
feat(req): add Accept header

### LINKED ISSUE #254: Include `Accept: application/yaml` header in GET requests for the driver index file.
In the future we're thinking of using content negotiation to serve the driver index in multiple formats (YAML, HTML). YAML will be the default (for compatibility) but adding this header can't hurt.

zeroshade:
closes #254

### DIFF
--- a/drivers.go
+++ b/drivers.go
@@ -159,7 +159,7 @@ func makereq(u string) (resp *http.Response, err error) {
 	}
 
 	if uri.Path == "/index.yaml" {
-		req.Header.Set("Accept", "application/x-yaml")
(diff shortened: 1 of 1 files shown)


drivers.go
```diff
@@ -158,6 +158,10 @@ func makereq(u string) (resp *http.Response, err error) {
 		Header: http.Header{},
 	}
 
+	if uri.Path == "/index.yaml" {
+		req.Header.Set("Accept", "application/x-yaml")
```
ianmcook:
```suggestion
		req.Header.Set("Accept", "application/yaml")
```

PR #257, opened 2026-01-20
fix(logout): add test for --purge

zeroshade:


ianmcook:
Later, if we have dbc add another file to the credentials path, but we forget to make the `--purge` implementation remove it, will this test fail?

zeroshade:
I can edit the test to validate that the directory is empty, but it would only fail if the additional file is added by `auth.AddCredential`

ianmcook:
Ok, I guess there's not really a good way to test for what I was imagining here, but this seems better than nothing. Thanks.

PR #265, opened 2026-01-21
feat(install): allow install from local package

### LINKED ISSUE #262: Idea: dbc install from local archive
This might be more useful for development and CI but it would be neat if dbc could install drivers from signed and unsigned release archives.

zeroshade:
closes #262

### DIFF
--- a/cmd/dbc/install.go
+++ b/cmd/dbc/install.go
@@ -168,7 +168,7 @@ type progressiveInstallModel struct {
 }
 
 func (m progressiveInstallModel) Init() tea.Cmd {
-	if strings.HasSuffix(m.Driver, ".tar.gz") {
+	if strings.HasSuffix(m.Driver, ".tar.gz") || strings.HasSuffix(m.Driver, ".tgz") {
 		return tea.Batch(m.spinner.Tick, func() tea.Msg {
 			return localInstallMsg{}
 		})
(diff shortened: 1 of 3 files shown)


cmd/dbc/install.go
```diff
@@ -162,9 +164,16 @@ type progressiveInstallModel struct {
 	p       dbc.FileProgressModel
 
 	width, height int
+	isLocal       bool
 }
 
 func (m progressiveInstallModel) Init() tea.Cmd {
+	if strings.HasSuffix(m.Driver, ".tar.gz") {
```
lidavidm: Should we also allow .tgz? Though it looks like CI only uses .tar.gz

zeroshade:
Yea, I don't think the info is there, the only option I have is to use the base name of the file. Otherwise we'd need to enforce the name of the package for us to parse it out? 

I guess I could just check if the filename matches that pattern?

lidavidm:
I think it should at least reject the package or maybe ask you to specify a name since the name as-is isn't usable by the driver manager.

zeroshade:
Why isn't the name usable by the driver manager? Do we mistake the extension and use the wrong code path? 

That said, it should be easy enough to add a parameter to specify a name. I'll do that

zeroshade:
@lidavidm Can you try and test this again? My updates should address the comments / issues you had.

lidavidm:
That does seem to do it

PR #267, opened 2026-01-21
feat: add a --proxy parameter

### LINKED ISSUE #243: Running dbc through proxy server
Are there any options to set a proxy server when running DBC with a corporate firewall?

pip supports a --proxy parameter.

uv  uses http_proxy and https_proxy environmental variables.

zeroshade:
closes #243


zeroshade:
Closing this for now. we seem to respect the HTTP_PROXY / HTTPS_PROXY parameters due to using the default transport so we don't need an explicit proxy parameter

PR #270, opened 2026-01-22
fix: update help text for dbc auth login

amoeba:
Updates the help text for `dbc auth login` to call out the default value. Uses a style like like other defaults args in the CLI.


amoeba:
Updated help text for `logout` to also specify there's a default (i.e., both default to the same value)

PR #279, opened 2026-01-28
fix(errors): refactor error handling

### LINKED ISSUE #268: Driver download progress bar remains visible after error
The download progress bar remains visible in the terminal even after an error has occurred and the error message is printed. See screenshot below.

Tested in Terminal on macOS.

Image

zeroshade:
Changes our error handling for the commands a bit to make it simpler and cleaner.

More specifically, this also fixes #268


amoeba (approved):
LGTM. And I was able to reproduce the old error by making my terminal narrower so I know this fixes it.

PR #282, opened 2026-01-29
fix: panic in search with invalid local manifests

amoeba:
Fixes a crash a user could encounter if they ran `dbc search` with an invalid driver manifest in a config dir. This can happen because when we run search we also enumerate all of the drivers on the users system so we have to load all the manifests. The fix here makes `decodeManifest` error if name and version aren't set which may also be fixing some latent bugs in other commands.

Closes https://github.com/columnar-tech/dbc/issues/281

### DIFF
--- a/config/driver_test.go
+++ b/config/driver_test.go
@@ -155,6 +155,7 @@ func TestLoadDriverFromInvalidManifest(t *testing.T) {
 			_, err := loadDriverFromManifest(prefix, driverName)
 			require.ErrorIs(t, err, ErrInvalidManifest)
 			require.ErrorContains(t, err, tt.errContains)
+			require.ErrorContains(t, err, manifestPath)
 		})
 	}
 }


config/config.go
```diff
@@ -299,6 +301,14 @@ func decodeManifest(r io.Reader, driverName string, requireShared bool) (Manifes
 			di.ManifestVersion, currentManifestVersion)
 	}
 
+	// Callers can assume these fields are set so return an error if they aren't
+	if di.Name == "" {
+		return Manifest{}, fmt.Errorf("%w: %s", ErrInvalidManifest, "name is required")
```
zeroshade: I wouldn't use %s for a constant, I would do `"%w: name is required"`

Same goes for below

amoeba: ah right, thanks

amoeba: done in cd875cca0c2307e45edd7e772f4bfc4d13a0b9c8

amoeba:
That's a good point. I double-checked where we call `decodeManifest` and the call site where I think we should print the path we already do. I added an assertion for that in e6549fee5b5f3090bd16855877e5997eb967682b. The other place we call `decodeManifest` I don't think we need to change it.

amoeba:
Merging.

Issue #285, opened 2026-02-02
Enable shell completions in Homebrew 


amoeba:
Homebrew formulae and casks can install shell completions at install-time and dbc has shell completions but we haven't enabled these in GoReleaser. We should.

- GoReleaser Docs: https://goreleaser.com/customization/homebrew_casks/
- Our tap: https://github.com/columnar-tech/homebrew-tap

Blocked on:

- [ ] Waiting for someone to submit dbc to homebrew/core (as a Formula, not Cask)

amoeba:
I think we have to hold off on this until dbc is in homebrew/core (as a formula, not cask). dbc's shell completions are generated dynamically and GoReleaser (and ultimately brew) needs these to be static files. See https://www.github.com/Homebrew/brew/issues/21482.

zeroshade:
Our completions *are* static files though (even though we output them dynamically). We could update the goreleaser config to publish the completions in the meantime while it's still a cask?

amoeba:
Oh, you're right. I don't know why I wasn't seeing them when I write this issue. Yeah we can totally do that.

JavOrraca:
Same issue (I think)... I am trying to install the dbc cask via homebrew but it appears to reference shell completion files that don't exist. Thanks folks!

amoeba:
Hey @JavOrraca, apologies for that. Can you try running `brew update` and installing again with `brew install dbc`? We pushed an update that fixes the installation though you'll have to install completions manually afterwards with `dbc completion`.

JavOrraca:
Late reply @amoeba but really appreciate the guidance, your recommendation to install completions manually afterwards worked.

PR #290, opened 2026-02-04
chore: update homebrew config

zeroshade:
Addresses #285 but doesn't close it as we'd still need to do different work if/when we convert to using a formula instead

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -316,10 +316,7 @@ homebrew_casks:
     commit_msg_template: "Brew cask update for {{ .ProjectName }} version {{ .Tag }}"
     homepage: "https://columnar.tech/dbc"
     description: "A CLI utility for managing ADBC drivers"
-    skip_upload: auto # auto skips prereleases
-    zap:
-      delete:
-        - ~/.config/columnar/dbc
+    skip_upload: auto # auto skips prereleases    
     completions:
       bash: cmd/dbc/completions/dbc.bash
       zsh: cmd/dbc/completions/dbc.zsh


.goreleaser.yaml
```diff
@@ -317,6 +317,13 @@ homebrew_casks:
     homepage: "https://columnar.tech/dbc"
     description: "A CLI utility for managing ADBC drivers"
     skip_upload: auto # auto skips prereleases
+    zap:
+      delete:
+        - ~/.config/columnar/dbc
```
amoeba: What's the idea behind this? I guess I have two comments,

1. This isn't the right dir is it?
2. We don't want to zap this since it contains credentials right?

zeroshade: that's the folder that will contain the uid.uuid (as far as I can tell)

zeroshade: credentials goes to `~/.local/share/dbc/credentials` etc....

Also, this would only be deleted if the user does `brew uninstall --zap`. If they don't explicitly ask to zap, then it would leave the file

amoeba: I thought all of this went in Application Support now?

```sh
$ tree ~/Library/Application\ Support/Columnar/dbc/
/Users/bryce/Library/Application Support/Columnar/dbc/
├── credentials
│   ├── columnar.lic
│   └── credentials.toml
└── uid.uuid
```

zeroshade: ... freaking macos

I guess i'll have to update this to be different for macos vs linux

zeroshade: I think I'll just remove the zap config for now. We can revisit this at a later time if users have feedback

PR #292, opened 2026-02-04
fix(config): use os.OpenRoot in uninstall

### LINKED ISSUE #55: Replace strings.HasPrefix with os.OpenRoot in driver uninstall
To ensure we are removing a directory that's safe to remove, we currently use a combination of filepath.Clean and strings.HasPrefix,

https://github.com/columnar-tech/dbc/blob/845f710596b7a31a10855a058ed1559a952d2bdd/config/config.go#L322-L329

Using https://go.dev/blog/osroot is probably even a better way to do this type of protection so we should refactor to use it.

This is a follow-on to https://github.com/columnar-tech/dbc/pull/48.

zeroshade:
closes #55


zeroshade:
ok, *now* this is ready for review @amoeba

PR #294, opened 2026-02-05
feat(prereleases): add `--pre` option to install, search, add

### LINKED ISSUE #289: Adding a `--pre` option for install/add/sync
Similar to the `--pre` option when using pip/uv we want to allow installing pre-releases of drivers via `dbc`. Both `uv` and `pip` have different behaviors for this so we should decide how we want to handle this.

`pip`'s behavior is very simplistic: Without the `--pre` option, pre-release versions of packages are never considered and essentially don't exist. The benefit here is that it's very simple to implement via filtering versions when reading package metadata.

`uv`'s behavior provides more options. Via either the `--prerelease` option or via a `UV_PRERELEASE` environment variable, one of the following options can be chosen to control the behavior:

- `disallow`: never allow pre-release versions (the same as `pip` without the `--pre` option)
- `allow`: allow pre-release versions, e.g. treat them as normal versions for consideration. This would mean that the "latest" version could be a pre-release (1.1.0-beta1 > 1.0.0)
- `if-necessary`: allow pre-release versions if a package *only* has pre-release versions, ignore them otherwise
- `explicit`: allow a pre-release version if explicitly asked for, ignore them otherwise
- `if-necessary-or-explicit`: allow a pre-release version if there are *only* pre-release versions for a package _or_ if the pre-release version is explicitly asked for

My personal preference is to follow `pip`'s behavior, but I'm not opposed to mirroring the behavior of `uv`. Thoughts?

zeroshade:
closes #289

To test with the staging registry, you need to use the `--no-verify` option as the current pre-release in the staging directory is signed with the nightly key, not the standard private key.

### DIFF
--- a/cmd/dbc/completions/dbc.bash
+++ b/cmd/dbc/completions/dbc.bash
@@ -69,7 +69,7 @@ _dbc_install_completions() {
     esac
 
     if [[ "$cur" == -* ]]; then
-        COMPREPLY=($(compgen -W "--json --no-verify --pre --level -l" -- "$cur"))
+        COMPREPLY=($(compgen -W "--json --no-verify --level -l --pre" -- "$cur"))
         return 0
     fi
 
--- a/cmd/dbc/completions/dbc.fish
+++ b/cmd/dbc/completions/dbc.fish
@@ -45,7 +45,7 @@ complete -f -c dbc -n '__fish_dbc_using_subcommand install' -s h -d 'Show Help'
 complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l help -d 'Show Help'
 complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l json -d 'Print output as JSON instead of plaintext'
 complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l no-verify -d 'Do not verify the driver after installation'
-complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l pre -d 'Include pre-release versions (hidden by default)'
+complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l pre -d 'Allow implicit installation of pre-release versions'
 complete -f -c dbc -n '__fish_dbc_using_subcommand install' -l level -s l -d 'Installation level' -xa 'user system'
 
 # uninstall subcommand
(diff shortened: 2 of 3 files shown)


drivers.go
```diff
@@ -451,21 +457,31 @@ func (d Driver) Versions(platformTuple string) semver.Collection {
...
+func (d Driver) GetPackage(version *semver.Version, platformTuple string, allowPrerelease bool) (PkgInfo, error) {
+	pkglist := d.PkgInfo
+	if !allowPrerelease && (version == nil || version.Prerelease() != "") {
+		pkglist = slices.Collect(filter(slices.Values(d.PkgInfo), func(p pkginfo) bool {
+			return p.Version.Prerelease() == ""
+		}))
+		if len(pkglist) == 0 {
+			return PkgInfo{}, fmt.Errorf("driver `%s` not found", d.Path)
+		}
+	}
```
lidavidm, line 469: I wonder if we could be nicer, and if there are no packages, we check if there are prereleases and say "driver not found (but prerelease versions filtered out)" 

We could even be extra-nice and run the version resolution to tell the user if a prerelease version would have fulfilled their criteria

lidavidm: Incidentally, this feels like the kind of thing where all the behaviors we want can be unit-tested directly on GetPackage?

ianmcook:
Looks like there's a bug in the **Available Versions** output in `dbc search --pre -v`.

ianmcook:
For the `search` subcommand, the `--help` description for `--pre` looks good.

But for the `install` and `add` subcommands, can you please change it to:
```
Allow implicit installation of pre-release versions
```

ianmcook:
When I explicitly specify a version constraint like `dbc install foo=v0.1.0-alpha.1`, dbc fails with this error when it should succeed:
```
Error: error getting driver: no package found for driver foo that satisfies constraints =v0.1.0-alpha.1
```

drivers.go
zeroshade: updated the message

ianmcook:
Update: the two apparent bugs that I described in the comments above were caused by an erroneous platform tuple on the registry index that I was testing with, not by errors in this PR.

cmd/dbc/add.go
```diff
@@ -44,20 +44,23 @@ func driverListPath(path string) (string, error) {
 type AddCmd struct {
 	Driver []string `arg:"positional,required" help:"Driver to add"`
 	Path   string   `arg:"-p" placeholder:"FILE" default:"./dbc.toml" help:"Driver list to add to"`
+	Pre    bool     `arg:"--pre" help:"Include pre-release versions (sets prerelease='allow')"`
```
ianmcook: I'd prefer this for `dbc add --help`. I don't think it's helpful to show `dbc.toml` syntax in the help.
```suggestion
	Pre    bool     `arg:"--pre" help:"Allow pre-release versions implicitly"`
```

PR #296, opened 2026-02-06
fix(search): provide graceful warning when a registry is offline

### LINKED ISSUE #295: Succeed with warning when private registry is offline
Currently, if `dbc-cdn-private` is offline or broken, `dbc search` commands fail.

Instead, let's make `dbc search` commands succeed with a warning. Of course the output will be limited to drivers from `dbc-cdn`, but that's better than a failure.

zeroshade:
fixes #295

Changes `dbc search` to succeed with a warning if any of the registries still succeed and return a list of drivers while one or more fail.

If there's an error but no drivers at all then we still fail entirely.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
-          go-version: 1.24.2
+          go-version: 1.25
           cache: true
           cache-dependency-path: go.sum
       - name: Build dbc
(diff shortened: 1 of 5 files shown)


ianmcook:
I didn't realize it when I opened the issue, but apparently `dbc install` and `dbc add` also fail in this case. Can you solve that too?

ianmcook:
If `dbc install` or `dbc add` finds the driver the user requests in the public registry, it shouldn't print a warning, it should just succeed.

ianmcook:
Edge case that this doesn't handle well: if BOTH registries are unreachable, it prints both a warning and an error:
image

zeroshade:
@ianmcook @amoeba updated based on the comments, please check again

ianmcook:
Looks like `dbc sync` suffers the same issue 🙃 
Can you fix that too please @zeroshade ?

ianmcook:
And `dbc info` and `dbc docs`, too!

zeroshade:
@ianmcook I've updated all the subcommands appropriately. we should have all the scenarios covered now

ianmcook:
Seems good now! Thanks @zeroshade

PR #303, opened 2026-02-08
chore: use the definite article to describe dbc

ianmcook:


amoeba (approved):
+1

Also updated goreleaser and the repo description on GH

Issue #309, opened 2026-02-10
Include completions files in release archives


amoeba:
When I try to install dbc from the most recent formula in our tap, I get an error about the completions and the install fails,

```sh
$ brew install dbc
==> Downloading https://github.com/columnar-tech/dbc/releases/download/v0.2.0/dbc-darwin-arm64-0.2.0.tar.gz
==> Downloading from https://release-assets.githubusercontent.com/github-production-release-asset/1052841794/37f82e3c-ab9f-43bd-8948-68d2ed22edbf?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-02-10T05%3A02%3A10Z&rscd=attachment%3B+filename%3Ddbc-darwin-arm64-0.2.0.tar.gz&rsct=application%2Foctet-stream&skoid=96c2
######################################################################################################################################################################################################################################################### 100.0%
==> Installing Cask dbc
==> Linking Binary 'dbc' to '/opt/homebrew/bin/dbc'
==> Unlinking Binary '/opt/homebrew/bin/dbc'
==> Purging files for version 0.2.0 of Cask dbc
Error: It seems the symlink source '/opt/homebrew/Caskroom/dbc/0.2.0/cmd/dbc/completions/dbc.bash' is not there.
```

It didn't occur to me when reviewing https://github.com/columnar-tech/dbc/pull/290  that we needed to actually have the completions available at install time.

PR #310, opened 2026-02-10
chore: include shell completions in archive, fix homebrew config

amoeba:
Closes https://github.com/columnar-tech/dbc/issues/309


amoeba:
This is my best guess at how this will work since I don't have a quick way to run goreleaser myself right now.

amoeba:
If we want to save time fixing this, we can just manually update the formula on our tap for tomorrow.

PR #320, opened 2026-02-17
fix: windows installer can't be uninstalled

### LINKED ISSUE #319: winget uninstall not fully uninstalling dbc and leaves system in inconsistent state
I had dbc 0.1.0 on my system via winget and when I uninstalled it with `winget uninstall`, it appears to succeed. But when I try to install dbc again with winget, it fails with this,

```
C:\Users\Bryce>winget install dbc
Found an existing package already installed. Trying to upgrade the installed package...
No applicable upgrade found.
A newer package version is available in a configured source, but it does not apply to your system or requirements.
```

When I check the registry (HKLM), I see dbc's uninstall is still registered but the path to its uninstall no longer exists. i.e., I think our uninstall doesn't remove itself from the registry like it should.

There's a workaround which is to `winget install --force dbc`.

amoeba:
Updates our WiX template to fix an issue where dbc can't be uninstalled with Add/Remove programs. We were setting two values in our .wxs template incorrectly. The "Product Code" should vary between releases and the "Package Code" should as well. 

See:

- https://learn.microsoft.com/en-us/windows/win32/msi/productcode
- https://learn.microsoft.com/en-us/windows/win32/msi/package-codes

Closes #319

### DIFF
--- a/resources/dbc.wxs
+++ b/resources/dbc.wxs
@@ -48,8 +48,8 @@ limitations under the License.
         <Publish Dialog="MaintenanceTypeDlg" Control="Back" Event="NewDialog" Value="MaintenanceWelcomeDlg">1</Publish>
     </UI>
   </Fragment>
-           Manufacturer="Columnar Technologies Inc." Version="{{ .Version }}"
   <Product Id="*" Name="dbc" Language="1033"
+           Manufacturer="Columnar Technologies Inc." Version="{{ .Version }}"
            UpgradeCode="3BC2470F-54EE-47F0-973C-A789684BF95E">
 
     <Package InstallScope="perUser"


amoeba:
I just downloaded the MSI from the CI artifacts and it installs and uninstalls (through add/remove programs) cleanly.

resources/dbc.wxs
```diff
@@ -48,12 +48,11 @@ limitations under the License.
         <Publish Dialog="MaintenanceTypeDlg" Control="Back" Event="NewDialog" Value="MaintenanceWelcomeDlg">1</Publish>
     </UI>
   </Fragment>
-  <Product Id="5B742771-61E1-4DC1-9276-6E1287282B3F" Name="dbc"
            Manufacturer="Columnar Technologies Inc." Version="{{ .Version }}"
+  <Product Id="*" Name="dbc" Language="1033"
```
zeroshade, line 51: I think this needs to be one line up

amoeba: Oh yep. Fixed in 8ed14020f408ee0896fc6b4a1fbb9daa7027125b.

amoeba: This isn't quite right. Will fix in a sec.

amoeba: Never mind, I can't read. Should be good now.

PR #328, opened 2026-03-05
fix(installing): use a '-' instead of check when using --no-verify

### LINKED ISSUE #305: Installing with --no-verify still uses a green checkmark for verifying step
```sh
$ go run ./cmd/dbc install --no-verify some_driver_unsigned_fixed.tar.gz
Installing from local package: some_driver_unsigned_fixed.tar.gz

[✓] installing
[✓] verifying signature

Installed some_driver_unsigned_fixed 1.10.0 to /Users/bryce/Library/Application Support/ADBC/Drivers
```

I think the `✓` should be a `-` or something to indicate the step was skipped.

zeroshade:
fixes #305

### DIFF
--- a/cmd/dbc/install.go
+++ b/cmd/dbc/install.go
@@ -415,10 +415,6 @@ func checkbox(label string, checked bool) string {
 
 var postMsgStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("3"))
 
-var (
-	skipMark = lipgloss.NewStyle().Foreground(lipgloss.Color("3")).SetString("-")
-)
-
 func (m progressiveInstallModel) View() string {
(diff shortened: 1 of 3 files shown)


cmd/dbc/install.go
```diff
@@ -415,6 +415,10 @@ func checkbox(label string, checked bool) string {
 
 var postMsgStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("3"))
 
+var (
+	skipMark = lipgloss.NewStyle().Foreground(lipgloss.Color("3")).SetString("-")
```
amoeba: Maybe merge this with the definition of `checkMark` in sync.go?

PR #329, opened 2026-03-05
chore: update to bubbletea v2

zeroshade:
Update to the recently released bubbletea v2 (and corresponding libs like lipgloss and such)

see https://github.com/charmbracelet/bubbletea/releases/tag/v2.0.0

### DIFF
--- a/cmd/dbc/main.go
+++ b/cmd/dbc/main.go
@@ -31,7 +31,9 @@ import (
 )
 
 var (
-	errStyle = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("1"))
+	errStyle  = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("1"))
+	checkMark = lipgloss.NewStyle().Foreground(lipgloss.Color("42")).SetString("✓")
+	skipMark  = lipgloss.NewStyle().Foreground(lipgloss.Color("3")).SetString("-")
 )
 
 type TuiCmd struct{}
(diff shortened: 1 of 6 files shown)


cmd/dbc/main.go
```diff
@@ -106,16 +113,16 @@ type baseModel struct {
...
 func (m baseModel) FinalOutput() string { return "" }
 
 func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
-	case tea.KeyMsg:
-		switch msg.Type {
-		case tea.KeyCtrlC, tea.KeyCtrlD, tea.KeyEsc:
+	case tea.KeyPressMsg:
+		switch msg.String() {
+		case "ctrl+c", "ctrl+d", "esc":
```
amoeba, line 127: This change in bubbletea seems unfortunate. Just noticing.

amoeba (commented):
I read through the https://github.com/charmbracelet/bubbletea/blob/v2.0.0/UPGRADE_GUIDE_V2.md and then reviewed each file. I then went through each file the changes here look pretty safe.

This does break piping output, I think something with the new NeedsRenderer method is off. For example, when I pipe search output, I now get control codes,

```
$ dbc search | less
ESC[1;38;5;35mbigqueryESC[m
```

Other than that, this looks good to me.

cmd/dbc/main.go
zeroshade: yea, that did annoy me a bit.

zeroshade:
@amoeba updated to fix the piping output

PR #330, opened 2026-03-09
fix: improve error when installing driver archive with subdirs

amoeba:
dbc currently can't install driver archives with subfolders but the error message the user gets could be improved. This isn't a major issue since we've been the ones doing all the packaging but others could run into this if they start creating driver archives. This change makes the reason for the error explicit.


zeroshade (approved):
Looks good!

PR #332, opened 2026-03-18
docs: improve install version constraint guidance

### LINKED ISSUE #331: Improve discoverability of version constraints for `dbc install`
I struggled to discover how to pass versions to `dbc install`.

What I eventually needed was already documented here:
- https://docs.columnar.tech/dbc/guides/installing/#version-constraints

But the pages I looked at first did not make that easy to find:
- https://docs.columnar.tech/dbc/ has install examples, but it does not mention that `dbc install` accepts version constraints or link to the detailed syntax docs.
- https://docs.columnar.tech/dbc/reference/cli/#install says:

  > DRIVER  
  > Name of the driver to install. Can be a short driver name or a driver name with version requirement. Examples: bigquery, bigquery=1.0.0, bigquery>1

  That hints at the feature, but it is easy to miss and does not point to the full version-constraint syntax.

The most confusing part was the CLI help itself. `dbc install -h` currently says only:

```text
Usage: dbc install [--level LEVEL] [--json] [--no-verify] [--pre] DRIVER

Positional arguments:
  DRIVER                 Driver to install
```

So if someone reaches for `dbc install -h`, there is no clue that inputs like `mysql=0.1.0` or `mysql>=1,<2` are valid.

I think this should be made more explicit in a few places:
- mention version constraints, with a link, from the docs home page
- make the CLI reference point directly to the version-constraints section
- make `dbc install -h` say that `DRIVER` can include a version constraint and show a couple of examples

dataders:
Closes #331.

## Summary
- make `dbc install -h` explicitly mention version constraints and link to the detailed docs
- add a regression test for the install help output
- add a visible version-constraint callout on the docs home page
- make the CLI reference point directly to the version-constraints guide

## Validation
- `go test ./...`
- `go run ./cmd/dbc install -h`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -75,8 +75,6 @@ dbc is the command-line tool for installing and managing [ADBC](https://arrow.ap
 
 <br/>2. Use dbc to install drivers:
 
-Need a specific version? See [Installing Drivers: Version Constraints](./guides/installing.md#version-constraints).
-
 === "BigQuery"
 
     ```console


docs/index.md
```diff
@@ -75,6 +75,8 @@ dbc is the command-line tool for installing and managing [ADBC](https://arrow.ap
 
 <br/>2. Use dbc to install drivers:
 
+Need a specific version? You can pass a version constraint directly to `dbc install`, for example `dbc install "bigquery>=1,<2"`. See [Installing Drivers: Version Constraints](./guides/installing.md#version-constraints).
```
dataders:
```suggestion
Need a specific version? See [Installing Drivers: Version Constraints](./guides/installing.md#version-constraints).
```

docs/index.md
```diff
@@ -75,6 +75,8 @@ dbc is the command-line tool for installing and managing [ADBC](https://arrow.ap
 
 <br/>2. Use dbc to install drivers:
 
+Need a specific version? See [Installing Drivers: Version Constraints](./guides/installing.md#version-constraints).
+
```
amoeba: The goal with this section on the index was to show a simple 1-2-3 so I hesitate to add extra text here. Do you think the overall benefit of your PR would be mostly in tact without this?

amoeba:
```suggestion
```

PR #334, opened 2026-03-27
feat: suggest common mistaken subcommands

amoeba:
This PR adds a way for dbc to suggest commands when the user enters a command
that doesn't exist. It adds a hook to our argument parser's error handler to
augment the error message with a suggestion.

I started with list -> search because I notice that users and/or AI agents like
to hallucinate 'dbc list' and I think this will help by at least reducing the
number of turns it takes for the agent to learn about dbc search.

We might consider a fuzzy suggestion feature (saerch -> search) like other CLI
tools use as a follow-on to this.

### DIFF
--- a/cmd/dbc/main.go
+++ b/cmd/dbc/main.go
@@ -190,8 +190,8 @@ func failSubcommandAndSuggest(p *arg.Parser, msg string, subcommand ...string) {
 		}
 	}
 
-	p.WriteUsageForSubcommand(os.Stderr, subcommand...)
-	fmt.Fprintf(os.Stderr, "error: %s", msg)
+	p.WriteUsageForSubcommand(os.Stdout, subcommand...)
+	fmt.Fprintf(os.Stdout, "error: %s", msg)
 
 	// Optionally add suggestion
 	if invalidCmd != "" {
@@ -231,7 +231,7 @@ func main() {
 	}
 
 	if p.Subcommand() == nil {
-		p.WriteHelp(os.Stderr)
+		p.WriteHelp(os.Stdout)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


amoeba:
Happy to take ideas on the look of this,

```sh
$ ./dbc list
Usage: dbc [--quiet] <command> [<args>]
error: invalid subcommand: list. Did you mean: dbc search?
```

cmd/dbc/main.go
```diff
@@ -172,6 +173,37 @@ func formatErr(err error) string {
...
+		for _, arg := range os.Args[1:] {
+			if !strings.HasPrefix(arg, "-") {
+				invalidCmd = arg
+				break
+			}
+		}
+	}
+
+	p.WriteUsageForSubcommand(os.Stdout, subcommand...)
+	fmt.Fprintf(os.Stdout, "error: %s", msg)
```
zeroshade, line 194: Should this go to `Stderr` instead?

cmd/dbc/main.go
```diff
@@ -172,6 +173,37 @@ func formatErr(err error) string {
...
+
+	// Optionally add suggestion
+	if invalidCmd != "" {
+		if suggestion, ok := subcommandSuggestions[invalidCmd]; ok {
+			fmt.Fprintf(os.Stdout, ". Did you mean: dbc %s?", suggestion)
```
zeroshade: Same question

cmd/dbc/main.go
```diff
@@ -172,6 +173,37 @@ func formatErr(err error) string {
...
+		}
+	}
+
+	fmt.Fprintln(os.Stdout)
```
zeroshade: Do we need the extra newline?

cmd/dbc/main.go
amoeba: Yeah, and I think we should fix the usage command too. I'll do that.

amoeba:
I'd like to put the change for printing help text from stdout to stderr in another PR so I addressed just the newline. 

@zeroshade did you have any thoughts on the look or wording?

PR #338, opened 2026-04-01
fix: improve install and add error messages for pre-release drivers

amoeba:
Updates `dbc install` and `dbc add` to provide command suggestions for pre-release drivers.

For example: Currently, if I try to dbc install a driver that only has a pre-release available, I get an error message that isn't that helpful,

```sh
$ dbc install fakedriver
Error: driver `fakedriver` not found (but prerelease versions filtered out)
```

This PR updates the error so it recommends the user try `--pre`:

```sh
$ dbc install fakedriver
Error: driver `fakedriver` not found (but prerelease versions filtered out); try: dbc install --pre fakedriver
```

This error only happens if the driver _only_ has pre-release versions and the user doesn't specify `--pre`. It's not a perfect heuristic but I think it'll work well.

### DIFF
--- a/cmd/dbc/add.go
+++ b/cmd/dbc/add.go
@@ -149,7 +149,14 @@ func (m addModel) Init() tea.Cmd {
 				}
 			} else {
 				if !m.Pre && !drv.HasNonPrerelease() {
-					err := fmt.Errorf("driver `%s` not found in driver registry index", spec.Name)
+					var err error
+					if len(drv.PkgInfo) > 0 {
+						// Has packages, but they're all prereleases
+						err = fmt.Errorf("driver `%s` not found in driver registry index (but prerelease versions filtered out); try: dbc add --pre %s", spec.Name, spec.Name)
+					} else {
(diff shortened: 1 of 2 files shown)


ianmcook:
Thanks for doing this!

amoeba:
Updated PR to include coverage for dbc add as well.

ianmcook:
What about the case when only pre-release versions are available on the registry, and you already have a pre-release version installed?

For example, what should happen when you do:

```sh
dbc install --pre singlestore
dbc install singlestore
```

zeroshade (approved):

zeroshade:
Bah, i missed @ianmcook's [question](https://github.com/columnar-tech/dbc/pull/338#issuecomment-4177829983).

@amoeba can you still answer the question there and determine if we need a follow-up PR?

amoeba:
I don't think that behavior is changed by this PR so I'll create a new issue.

amoeba:
https://github.com/columnar-tech/dbc/issues/340

PR #346, opened 2026-04-06
chore: move pre-commit to separate workflow

amoeba:
Moves the recent pre-commit workflow addition to a new workflow file because
it needs to be run under pull_request and not pull_request_target.


amoeba:
Just needs a rebase after https://github.com/columnar-tech/dbc/pull/344.

zeroshade (approved):

PR #352, opened 2026-04-10
feat: dbc auth license install

amoeba:
I think we could make the steps in https://docs.columnar.tech/dbc/guides/private_drivers/#downloading-your-license even simpler if dbc handled putting your license in the right place for the user. This PR is a proposal for adding a `dbc auth license install` sub-subcommand to handle this. I tucked it under `dbc auth`.

### DIFF
--- a/auth/credentials.go
+++ b/auth/credentials.go
@@ -262,9 +262,53 @@ func IsColumnarPrivateRegistry(u *url.URL) bool {
 }
 
 var (
-	ErrNoTrialLicense = errors.New("no trial license found")
-	ErrTrialExpired   = errors.New("trial license has expired")
+	ErrNoTrialLicense       = errors.New("no trial license found")
+	ErrTrialExpired         = errors.New("trial license has expired")
+	ErrLicenseWrongFilename = errors.New("source file is not named columnar.lic (use --force to override)")
+	ErrLicenseAlreadyExists = errors.New("license already exists (use --force to overwrite)")
 )
+
+func LicensePath() string {
+	return filepath.Join(filepath.Dir(credPath), "columnar.lic")
+}
+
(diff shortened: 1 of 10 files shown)


amoeba:
(I still need to review this with fresh eyes tomorrow)

auth/credentials.go
```diff
@@ -262,10 +262,54 @@ func IsColumnarPrivateRegistry(u *url.URL) bool {
...
+func InstallLicenseFromFile(srcPath string, force bool) error {
+	src, err := os.Open(srcPath)
+	if err != nil {
+		return fmt.Errorf("failed to read license file: %w", err)
+	}
+	defer src.Close()
+
+	if !force && filepath.Base(srcPath) != "columnar.lic" {
+		return ErrLicenseWrongFilename
+	}
```
zeroshade: Couldn't we just use `os.ReadFile` and `os.WriteFile`? Seems like it would be simpler.

Something along the line of:

```go
destPath := LicensePath()
if !force {
        // check with os.Stat
}

// os.MkdirAll....

src, err := os.ReadFile(srcPath)
if err != nil {
        return err
}

return os.WriteFile(destPath, src, 0o700)
```

amoeba: yeah that's fine. Simplified in b7d1d65ade50fd56e2e3ef9e013d8e8bf2bb4ed0.

zeroshade (approved):
Looks good to me, anything else to make it ready for review rather than just a draft?

amoeba:
That's it. Thanks for the review, I'll merge this.

amoeba:
I should have pulled the docs changes out into a separate PR so we can update the docs separately so I filed a PR to revert those: https://github.com/columnar-tech/dbc/pull/358.

PR #354, opened 2026-04-10
fix: fix rendering bubbletea v2 regressions for local install

### LINKED ISSUE #351: dbc install from local package gives garbled output
With dbc from main

```
$ go run ./cmd/dbc install sqlite_macos_amd64_v1.10.0.tar.gz
Installing from local package: sqlite_macos_amd64_v1.10.0.tar.gz

[✓] installing
[✓] verifying signature

Installed sqlite_macos_amd64_v1.10.0 1.10.0 to /Users/bryce/Library/Application Support/ADBC/Drivers
^[[?2026;2$y^[[?2027;2$y⏎
```

old behavior (dbc 0.2.0, using old bubbletea v1 renderer):

```
$ dbc install sqlite_macos_amd64_v1.10.0.tar.gz
Installing from local package: sqlite_macos_amd64_v1.10.0.tar.gz

[✓] installing
[✓] verifying signature

Installed sqlite_macos_amd64_v1.10.0 1.10.0 to /Users/bryce/Library/Application Support/ADBC/Drivers
```

I'm assuming this is due entirely to the bubbletea v2 upgrade.

amoeba:
This fixes two regressions in bubbletea v2,

1. tea.Println inside Update() breaks scrollback
2. If a subcommand with a TUI (i.e., install) returns too quickly, bubbletea runs into a timing issue between switching between cooked and raw mode. I think bubbletea starts in cooked mode, switches to raw mode while showing the TUI (i.e., during prog.Run()), and switches back to cooked mode after. prog.Run() starts and stops so quickly when we run dbc install on local archive that some terminals haven't sent their terminal probe response before bubbletea switches back into cooked mode so the probe response shows up.

The fix for (2) was LLM written so a review there would be good.

This PR is split into separate commits so when you review this, you'll need to find a terminal that reproduces the issue (Ghostty, iTerm2, WezTerm I think too) and then work through the commits one by one to see the fixes. Issues (1) and (2) above are fixed by separate standalone commits.

Closes #351

### DIFF
--- a/cmd/dbc/drain_unix.go
+++ b/cmd/dbc/drain_unix.go
@@ -1,0 +1,75 @@
+// Copyright 2026 Columnar Technologies Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !windows
+
+package main
+
+import (
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/charmbracelet/x/term"
+)
+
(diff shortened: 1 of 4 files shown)


amoeba:
Hey @zeroshade, could you take a look at this, specifically if you can reproduce the original two issues and verify each of the fixes.

zeroshade (approved):
I was able to reproduce the issues and verified the fixes. Nice! Thanks!

PR #356, opened 2026-04-13
docs: create guides for version control and continuous integration

### LINKED ISSUE #164: Create a guide showing how to use dbc with Git
Add a guide that shows how to use dbc in the context of a collaborative project in Git.

Points to cover:
 
- When committing to a repo or creating a new project that will go into Git:
  - Use `dbc init`, `dbc add`, and `dbc sync` instead of `dbc install`.
  - Add `dbc.toml` to the repo.
  - Don't add the drivers to the repo
- After cloning a repo:
  - Run `dbc sync`

amoeba:
Creates guides for using dbc in version control and with CI. Incidentally syncs up some of the index.md files and mkdocs.yml since they were slightly out of sync.

Closes #164
Closes #342

### DIFF
--- a/docs/guides/continuous_integration.md
+++ b/docs/guides/continuous_integration.md
@@ -20,7 +20,7 @@ dbc works well in non-interactive environments such as on continuous integration
 
 ## GitHub Actions
 
-We recommend using the [columnar-tech/setup-dbc](https://github.com/columanr-tech/setup-dbc) action if you're using [GitHub Actions](https://docs.github.com/en/actions) for CI.
+We recommend using the [columnar-tech/setup-dbc](https://github.com/columnar-tech/setup-dbc) action if you're using [GitHub Actions](https://docs.github.com/en/actions) for CI.
 
 As an example, here's a workflow that automatically installs all drivers listed in your [driver list](../concepts/driver_list.md) before running your tests:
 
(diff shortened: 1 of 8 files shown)


docs/guides/continuous_integration.md
```diff
@@ -0,0 +1,92 @@
...
+on: [push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      # Note: Automatically installs drivers specified in dbc.toml
+      - uses: columnar-tech/setup-dbc@v1
```
zeroshade, line 38: do we want to add an example for simply installing individual drivers instead of the dbc.toml etc.?

zeroshade (approved):

docs/guides/continuous_integration.md
amoeba: Maybe? When I wrote this, I was thinking that we should nudge users to use driver lists wherever we can. I personally didn't like specifying drivers inline and I'm not sure users will like it either.

zeroshade: If we're not going to include the example, can we at least make sure we point at the setup-dbc README which has all the possible options and examples?

amoeba: I think this PR currently does that in a way that will work well.

Screenshot 2026-04-15 at 4 13 36 PM

Thoughts on that?

zeroshade: looks good!

PR #357, opened 2026-04-15
docs: update docs and test behavior for --pre

### LINKED ISSUE #327: Document and test behavior of `dbc search` without `--pre` when driver is already installed
I was momentarily confused today when I ran `dbc search` without `--pre` and it showed the ClickHouse driver, since that driver is only available as a pre-release.

Turns out the reason it was showing it is that I had already installed it, and `dbc search` will always list installed drivers regardless of pre-release status.

I think this behavior makes sense, but it's not documented anywhere, and I'm not sure if it's tested. We should document it and add tests if there aren't tests already.

Documenting it could be as simple as changing `(hidden by default)` to `(hidden by default unless already installed)` in a few places.

zeroshade:
fixes #327


ianmcook (approved):

amoeba (approved):
Looks good, thanks.

I added one more test to cover non-verbose search in 9d6d67b439098ad0207bafc2c73315a7af221d32.

PR #360, opened 2026-04-17
feat(registry): configurable driver registries via TOML config

zeroshade:
## Summary

Adds support for configuring driver registries via TOML config files, allowing users to add custom registries at both the global (`~/.config/columnar/dbc/config.toml`) and project (`dbc.toml`) level.

## What's New

- **Global registry config** (`~/.config/columnar/dbc/config.toml`): loaded on every command via `ConfigureRegistries()` called in `main()`
- **Project registry config** (`dbc.toml`): `[[registries]]` section respected by `sync` and `add` commands
- **Priority order**: project → global → built-in defaults (first-match-wins, URL-deduplicated)
- **`replace_defaults = true`**: opt-in flag to suppress built-in registries entirely
- **`DBC_BASE_URL` env var**: still overrides everything (unchanged behavior)
- **Backward compatible**: existing `dbc.toml` files without `[[registries]]` work unchanged

## Config Format

**Global** (`~/.config/columnar/dbc/config.toml`):
```toml
[[registries]]
url = "https://my-registry.example.com"
name = "my-registry"

# replace_defaults = true  # omit built-in registries
```

**Project** (`dbc.toml`):
```toml
[[registries]]
url = "https://project-registry.example.com"
name = "project"

[drivers]
[drivers.my-driver]
version = ">=1.0.0"
```

## Implementation Notes

- URL validation requires `http`/`https` scheme and non-empty host
- `defaultRegistries` is snapshotted eagerly in `init()` (after `drivers.go` init runs); env-sensitivity documented in comments
- `add.go` decodes `dbc.toml` once (single file open) for both registry config and driver list
- `remove` intentionally omits registry wiring — it never queries a registry


registry_config.go
```diff
@@ -0,0 +1,214 @@
...
+		registries = append([]Registry(nil), defaultRegistries...)
+		return nil
+	}
+	registries = mergeRegistries(nil, nil, cfg.Registries, cfg.ReplaceDefaults, defaultRegistries)
+	return nil
+}
+
+func SetProjectRegistries(entries []RegistryEntry, replaceDefaults *bool) error {
+	if os.Getenv("DBC_BASE_URL") != "" {
+		return nil
```
amoeba: I think we should warn the user here.

registry_config.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	if replaceDefaults != nil && *replaceDefaults && len(entries) == 0 {
+		return fmt.Errorf("replace_defaults = true requires at least one [[registries]] entry; omit replace_defaults or add a registry entry")
```
amoeba: I feel like this error would be better surfaced earlier, such as when we initially parse config.toml or dbc.toml. How this is now, I think this would error like this,

> error configuring project registries: replace_defaults = true requires at least one [[registries]] entry; omit replace_defaults or add a registry entry

As a user, it's unclear what I should do to fix it. Which file is my problem in?

If we validated this earlier, I bet we'd have enough context to know which file has the issue and we could tell them.

registry_config.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	for _, e := range entries {
+		if e.URL == "" {
+			return fmt.Errorf("registry entry has empty url")
```
amoeba: Are these checks needed or is this the best place to handle this? I see `loadGlobalConfig` already does validation and that seems like a better place. i.e., I much prefer being able to know that any `[]RegistryEntry` I might get is always valid so I'd rather error before creating a `RegistryEntry`.

cmd/dbc/main.go
```diff
@@ -103,6 +104,8 @@ func downloadPkg(p dbc.PkgInfo) (*os.File, error) {
 	})
 }
 
+func boolPtr(b bool) *bool { return &b }
```
amoeba: is this the best place for this def?

cmd/dbc/main.go
```diff
@@ -244,6 +247,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	if configDir, err := internal.GetUserConfigPath(); err != nil {
+		fmt.Fprintf(os.Stderr, "warning: failed to locate config directory: %v\n", err)
```
amoeba: Do we really want to emit a warning? Won't this warn if the folder isn't found? Seems like we should be okay with that.

The warning when the registry config fails to load once found is fine.

.gitignore
```diff
@@ -49,3 +49,6 @@ cdn-dev
 .claude
 
 __pycache__
+
+# Sisyphus workflow artifacts (plans, evidence, notepads)
+.sisyphus/
```
amoeba: is this just a tool you use? can we revert this?

cmd/dbc/main.go
```diff
@@ -244,6 +247,12 @@ func main() {
...
+	} else if err := dbc.ConfigureRegistries(configDir); err != nil {
+		fmt.Fprintf(os.Stderr, "warning: failed to load registry config: %v\n", err)
```
amoeba: I don't think the error here includes the file with the issue. Whenever we show an error about a TOML file, we need to show the user the path to it.

registry_config.go
```diff
@@ -0,0 +1,214 @@
...
+		return u.Scheme + "://" + u.Host + path
+	}
+
+	toRegistry := func(entry RegistryEntry) (Registry, bool) {
+		u, err := url.Parse(entry.URL)
+		if err != nil || u.Host == "" {
+			return Registry{}, false
+		}
+		if u.Scheme != "http" && u.Scheme != "https" {
+			return Registry{}, false
```
amoeba: Is the validation in here even needed? In the parent call, we already error when a RegistryEntry is invalid.

amoeba (commented):
I did a quick code review over the new pieces and left some comments. Even if we don't have an immediate use case for this, I think having it in dbc is useful.

Some other more general thoughts:

1. Would we want CLI integration for this? I think it would be useful for dbc to be able to do all the work to show/add/remove registries from the project and user config.
2. I didn't see any tests for how `dbc install` will work with this. Presumably dbc install will respect the new fields in the user config.
3. Do we need to define a default priority order for registries to handle name conflicts? We have a lot of the same concerns as conda does with channels. I'm wishing for more tests in this PR for the interaction between the default, user, and project registries around name conflicts.

zeroshade:
> Would we want CLI integration for this? I think it would be useful for dbc to be able to do all the work to show/add/remove registries from the project and user config.

Give the size of this PR, we can add CLI integration as a follow-up PR.

> I didn't see any tests for how dbc install will work with this. Presumably dbc install will respect the new fields in the user config.

Good point, i'll add tests for this

> Do we need to define a default priority order for registries to handle name conflicts? We have a lot of the same concerns as conda does with channels. I'm wishing for more tests in this PR for the interaction between the default, user, and project registries around name conflicts.

`dbc search` would show all the driver matches along with their `[registry]` name tags in the output. The priority order would be:

1. Project (dbc.toml)
2.  Global (~/.config/columnar/dbc/config.toml)
3. built-in defaults

Having `replace_defaults = true` at either the Project or Global config level would drop the defaults, but doesn't change the priority that registries at the project level would win over the global config level.  

That said, there's no tests currently for that behavior so I'll add some tests for it.

registry_config.go
zeroshade: Refactored away in 3068390 — `SetProjectRegistries` was replaced by the option-based `WithProjectRegistries` / `WithGlobalConfig` / `WithBaseURL` API on `NewClient`. The DBC_BASE_URL precedence is now handled by the `WithBaseURL` option (whose doc comment notes it overrides `WithGlobalConfig`/`WithProjectRegistries`), so this short-circuit code path no longer exists.

registry_config.go
zeroshade: Addressed in 6041d4a ("tri-state replace_defaults, validate global config"): the `replace_defaults=true` + zero-entries check moved into `LoadGlobalConfig` and `NewClient`'s post-merge guard, with file-aware error messages from `LoadGlobalConfig` (e.g. `registry entry in /path/to/config.toml has empty url`).

Followed up by c238087 ("defer replace_defaults zero-entry check to post-merge") which moves the empty-list check to after merging — so the legitimate case of `global replace_defaults=true` + project supplies entries is no longer rejected, and the error only fires when the merged list is genuinely empty.

registry_config.go
zeroshade: Consolidated in 3068390 ("add configurable registry options to Client"). The duplicated checks in `SetProjectRegistries` are gone — validation runs in a single place inside `mergeRegistries.addEntries` (registry_config.go), with `LoadGlobalConfig` performing the file-aware sanity check up front. `RegistryEntry` is now treated as the raw input shape and validated once at merge time.

cmd/dbc/main.go
zeroshade: Removed from `cmd/dbc/main.go` in 6041d4a. The helper now lives in `registry_config_test.go` (the only place it's still used, for setting tri-state `*bool` test inputs).

cmd/dbc/main.go
zeroshade: Addressed in beb7ace ("skip global config load when config dir unresolvable"): when `GetUserConfigPath` fails, `runStartup` now silently skips loading the global config rather than warning. The warning when the file is *found but fails to load* remains, which matches what you suggested.

.gitignore
zeroshade: Reverted — the `.sisyphus/` entry was rebased out and the current branch's `.gitignore` is identical to `main` (`git diff main..HEAD -- .gitignore` is empty).

cmd/dbc/main.go
zeroshade: Addressed in 6041d4a — errors from `LoadGlobalConfig` now embed the full `configPath`, e.g.:

```
invalid registry URL "..." in /home/user/.config/columnar/dbc/config.toml: scheme must be http or https
```

The CLI surfaces these errors verbatim, so the user always knows which TOML file is the source.

registry_config.go
zeroshade: Consolidated in 3068390. The `toRegistry` helper is gone — validation lives in a single place inside `mergeRegistries.addEntries`, and `LoadGlobalConfig` runs the same scheme/host/empty checks at parse time with file-aware error messages.

zeroshade:
@amoeba I addressed the comments, updated this, and added the tests that were mentioned. Let me know what you think

zeroshade:
the failing snapshot CI is an upstream snapd problem

amoeba:
I pushed a tiny safeguard in https://github.com/columnar-tech/dbc/pull/360/commits/7cb0e41e907d781af86b5d34863b438f45a8e465 since we now have user-injectable registry labels. Feel free to tweak.

I think the last thing I'm not sure I like is that `dbc search` isn't aware of custom registries defined in dbc.toml which I think would lead to confusion. If a user has a custom registry set in dbc.toml,

```
$ cat dbc.toml
replace_defaults = true

[[registries]]
url = 'http://localhost:8080'
name = 'local-fake'
```

and they try to dbc add it,

```
$ go run ./cmd/dbc add f
Error: driver `f` not found in driver registry index; try: `dbc search` to list available drivers
exit status 1
```

we suggest 'dbc search' but dbc search won't list drivers in the `local-fake` repo because it's only been set in dbc.toml and not in the global config. I also note wonder if `replace_defaults` in dbc.toml should effect dbc search and not just add/sync.

Is there any downside to simplifying the logic so the behavior is more consistent between the project-level and global dbc config?

zeroshade:
Can't think of a downside offhand, I agree that the behavior should be more consistent. I'll fix that

zeroshade:
@amoeba good call — addressed in 59a3514 (+ a follow-up in 2171cdf).

`dbc search`, `dbc info`, and `dbc docs` now load `./dbc.toml` and apply its `[[registries]]` + `replace_defaults` before querying, so they resolve drivers against the same registry set that `dbc add`/`dbc sync` use in the same project. Your example now works: with a project `dbc.toml` setting `replace_defaults = true` and a `local-fake` registry, `dbc search` lists that registry's drivers, and the `dbc add` "not found → try `dbc search`" suggestion is coherent. `replace_defaults` in `dbc.toml` affects these commands too now.

A missing `dbc.toml` is a no-op (these commands still work outside a project), and a malformed one is a hard error that names the file.

On scope: I deliberately left `dbc install` reading only the global + built-in default registries — it installs into a user/system [config level](https://docs.columnar.tech/dbc/reference/config_level) rather than the project, so making it silently cwd-sensitive felt like a surprising, larger change than this. Happy to revisit if you'd prefer install to be project-aware too.

One nuance worth flagging: `DBC_BASE_URL` still overrides everything and short-circuits `dbc.toml` entirely, so a broken project file can't block that escape hatch (2171cdf).

Docs updated in `docs/concepts/driver_registry.md` and `docs/reference/driver_list.md`.

amoeba (approved):
I just checked the final thing I noted was addressed. Can we back out the docs changes before merge and do the docs updates in a separate pr?

zeroshade:
Backed out the docs changes here (`docs/concepts/driver_registry.md` and `docs/reference/driver_list.md`) in 6012dbd and moved them to a separate docs PR: #393. This PR now contains only the implementation + tests.

PR #369, opened 2026-04-30
feat: add `list` subcommand to list installed drivers

zeroshade:
Adding a subcommand to list the currently installed drivers, along with updating the completion scripts and the docs for it.

### DIFF
--- a/cmd/dbc/list.go
+++ b/cmd/dbc/list.go
@@ -1,0 +1,197 @@
+// Copyright 2026 Columnar Technologies Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
+	"charm.land/lipgloss/v2/table"
+	"github.com/columnar-tech/dbc/config"
+	"github.com/columnar-tech/dbc/internal/jsonschema"
+)
+
+type ListCmd struct {
+	Level config.ConfigLevel `arg:"-l" help:"Only list drivers installed at this config level (user, system)"`
+	Json  bool               `arg:"--json" help:"Print output as JSON instead of plaintext"`
+}
+
+func (ListCmd) Description() string {
+	return "List all currently installed drivers across user, system, and environment config levels."
+}
+
+func (c ListCmd) GetModel() tea.Model {
+	return listModel{
+		level:      c.Level,
+		jsonOutput: c.Json,
+	}
+}
+
+type installedDriver struct {
+	Level   config.ConfigLevel
+	ID      string
+	Name    string
+	Version string
+	Path    string
+}
+
+type installedDriversMsg []installedDriver
+
+type listModel struct {
+	baseModel
+
+	level      config.ConfigLevel
+	jsonOutput bool
+	drivers    []installedDriver
+}
+
+func (m listModel) Init() tea.Cmd {
+	return func() tea.Msg {
+		cfgs := config.Get()
+
+		var levels []config.ConfigLevel
+		if m.level == config.ConfigUnknown {
+			levels = []config.ConfigLevel{config.ConfigSystem, config.ConfigUser, config.ConfigEnv}
+		} else {
+			levels = []config.ConfigLevel{m.level}
+		}
+
+		var drivers []installedDriver
+		for _, lvl := range levels {
+			cfg, ok := cfgs[lvl]
+			if !ok {
+				continue
+			}
+			for _, d := range cfg.Drivers {
+				version := ""
+				if d.Version != nil {
+					version = d.Version.String()
+				}
+				drivers = append(drivers, installedDriver{
+					Level:   lvl,
+					ID:      d.ID,
+					Name:    d.Name,
+					Version: version,
+					Path:    d.FilePath,
+				})
+			}
+		}
+
+		sort.Slice(drivers, func(i, j int) bool {
+			if drivers[i].Level != drivers[j].Level {
+				return drivers[i].Level < drivers[j].Level
+			}
+			return drivers[i].ID < drivers[j].ID
+		})
+
+		return installedDriversMsg(drivers)
+	}
+}
+
+func (m listModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case installedDriversMsg:
+		m.drivers = []installedDriver(msg)
+		return m, tea.Quit
+	default:
+		bm, cmd := m.baseModel.Update(msg)
+		m.baseModel = bm.(baseModel)
+		return m, cmd
+	}
+}
+
+func (m listModel) View() tea.View { return tea.NewView("") }
+
+func (m listModel) IsJSONMode() bool { return m.jsonOutput }
+
+func (m listModel) FinalOutput() string {
+	if m.status != 0 {
+		if m.jsonOutput {
+			return marshalEnvelope("error", jsonschema.ErrorResponse{
+				Code:    "list_failed",
+				Message: m.err.Error(),
+			})
+		}
+		return ""
+	}
+
+	if m.jsonOutput {
+		return listDriversJSON(m.drivers)
+	}
+	return formatInstalledDrivers(m.drivers)
+}
+
+func formatInstalledDrivers(drivers []installedDriver) string {
+	if len(drivers) == 0 {
+		return "No drivers installed."
+	}
+
+	t := table.New().Border(lipgloss.HiddenBorder()).
+		BorderTop(false).BorderBottom(false).BorderLeft(false).BorderRight(false).
+		Headers("DRIVER", "VERSION", "LEVEL", "LOCATION")
+	headerStyle := lipgloss.NewStyle().Bold(true)
+	levelStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("63"))
+	versionStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("42"))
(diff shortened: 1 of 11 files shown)


amoeba (commented):
Thanks for taking this on.

I ran this once and got,

```
$ go run ./cmd/dbc list
DRIVER    VERSION LEVEL LOCATION

duckdb    1.5.2   user  /Users/bryce/Library/Application Support/ADBC/Drivers
flightsql 1.11.0  user  /Users/bryce/Library/Application Support/ADBC/Drivers
snowflake 1.10.3  user  /Users/bryce/Library/Application Support/ADBC/Drivers
sqlite    1.11.0  user  /Users/bryce/Library/Application Support/ADBC/Drivers
```

I then ran

```
$ sudo dbc install --level system sqlite
```

and then the list command again list seems to forget I have sqlite installed at the user level:

```
$ go run ./cmd/dbc list
DRIVER    VERSION LEVEL  LOCATION

sqlite    1.11.0  system /Library/Application Support/ADBC/Drivers
duckdb    1.5.2   user   /Users/bryce/Library/Application Support/ADBC/Drivers
flightsql 1.11.0  user   /Users/bryce/Library/Application Support/ADBC/Drivers
snowflake 1.10.3  user   /Users/bryce/Library/Application Support/ADBC/Drivers
sqlite    1.11.0  user   /Users/bryce/Library/Application Support/ADBC/Drivers
```

Can `dbc list` just do this by default?

```
$ go run ./cmd/dbc list
sqlite
duckdb
flightsql
snowflake
sqlite
```

cmd/dbc/list.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if m.jsonOutput {
+		return listDriversJSON(m.drivers)
+	}
+	return formatInstalledDrivers(m.drivers)
+}
+
+func formatInstalledDrivers(drivers []installedDriver) string {
+	if len(drivers) == 0 {
+		return "No drivers installed."
```
ianmcook: Should this go to stderr instead of stdout?

ianmcook: The logic of using stderr is:

stdout is the data channel, stderr is the human channel. "No drivers installed" is a note about the data, not the data itself, so it belongs on stderr, `otherwise dbc list -q | wc -l` lies.

ianmcook:
> seems to forget I have sqlite installed at the user level

it's still there, on a separate line, below, isn't it?

> Can dbc list just do this by default?

I'm -1 on this. I think the current output is good, but I can see a case for maybe adding a `--quiet, -q` option that prints the output like you suggest (just the slugs, one per line). But if we add that, we should add it to the `search` subcommand too I think.

EDIT: Oops, it can't be `--quiet, -q` because we're already using that globally to suppress all output. (In retrospect, I guess `--silent` would have been a better choice for that one.) But we could use something else like `--names`.

amoeba:
> it's still there, on a separate line, below, isn't it?

Ah, I missed that. Interesting.

cmd/dbc/list.go
amoeba: I'm actually not sure we need this output at all.

cmd/dbc/list.go
```diff
@@ -0,0 +1,200 @@
...
+
+	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
+	"charm.land/lipgloss/v2/table"
+	"github.com/columnar-tech/dbc/config"
+	"github.com/columnar-tech/dbc/internal/jsonschema"
+)
+
+type ListCmd struct {
+	Level config.ConfigLevel `arg:"-l" help:"Only list drivers installed at this config level (user, system)"`
```
amoeba, line 32: Hrm, in `install`, we nixed `--level env` but we now need it with `list` don't we?

cmd/dbc/list.go
```diff
@@ -0,0 +1,200 @@
...
+	Json  bool               `arg:"--json" help:"Print output as JSON instead of plaintext"`
+}
+
+func (ListCmd) Description() string {
+	return "List all currently installed drivers across user, system, and environment config levels."
```
amoeba:
```suggestion
	return "List installed drivers across user, system, and environment config levels."
```

I think this could be simpler.

cmd/dbc/list.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		var drivers []installedDriver
+		for _, lvl := range levels {
+			cfg, ok := cfgs[lvl]
+			if !ok {
+				continue
+			}
+			if cfg.Err != nil {
+				return fmt.Errorf("failed to load drivers at %s level: %w", lvl, cfg.Err)
```
amoeba:
```suggestion
				return fmt.Errorf("failed to list drivers at %s level: %w", lvl, cfg.Err)
```

cmd/dbc/list.go
```diff
@@ -0,0 +1,200 @@
...
+					Name:    d.Name,
+					Version: version,
+					Path:    d.FilePath,
+				})
+			}
+		}
+
+		sort.Slice(drivers, func(i, j int) bool {
+			if drivers[i].Level != drivers[j].Level {
+				return drivers[i].Level < drivers[j].Level
```
amoeba:
```suggestion
				return drivers[i].Level > drivers[j].Level
```

env > user > system?

cmd/dbc/list_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	unreadable := filepath.Join(suite.T().TempDir(), "unreadable")
+	suite.Require().NoError(os.Mkdir(unreadable, 0o000))
+	suite.T().Cleanup(func() { _ = os.Chmod(unreadable, 0o700) })
+	suite.T().Setenv("ADBC_DRIVER_PATH", filepath.Join(unreadable, "drivers"))
+
+	m := ListCmd{Level: config.ConfigEnv}.GetModel()
+	out := suite.runCmdErr(m)
+	suite.Contains(out, "failed to load drivers")
```
amoeba:
```suggestion
	suite.Contains(out, "failed to list drivers")
```

and other fixes if we update the error

cmd/dbc/main.go
```diff
@@ -168,6 +168,7 @@ func (m baseModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 type cmds struct {
 	Install    *InstallCmd      `arg:"subcommand" help:"Install a driver"`
 	Uninstall  *UninstallCmd    `arg:"subcommand" help:"Uninstall a driver"`
+	List       *ListCmd         `arg:"subcommand" help:"List all currently installed drivers"`
```
amoeba:
```suggestion
	List       *ListCmd         `arg:"subcommand" help:"List installed drivers"`
```

docs/reference/cli.md
```diff
@@ -40,6 +40,7 @@ $ dbc [OPTIONS] <COMMAND>
 <dt><a href="#search">dbc search</a></dt><dd><p>Search for a driver to install</p></dd>
 <dt><a href="#install">dbc install</a></dt><dd><p>Install a driver</p></dd>
 <dt><a href="#uninstall">dbc uninstall</a></dt><dd><p>Uninstall a driver</p></dd>
+<dt><a href="#list">dbc list</a></dt><dd><p>List all currently installed drivers</p></dd>
```
amoeba:
```suggestion
<dt><a href="#list">dbc list</a></dt><dd><p>List installed drivers</p></dd>
```

amoeba: + changes elsewhere below

docs/reference/cli.md
```diff
@@ -155,6 +156,32 @@ $ dbc uninstall [OPTIONS] <DRIVER>
 
 :   Suppress all output
 
+## list
+
+{{ since_version('v0.3.0') }}
+
+List all currently installed drivers across every [configuration level](config_level.md) (system, user, and environment). Use `--level` to restrict the output to drivers installed at a single level.
```
amoeba: We haven't told the user about the "environment" config level so I don't think we can talk about it quite like this.  I think language like this needs to be more like "user, system or from environment variables"

cmd/dbc/list.go
ianmcook: I think we need it. Users will be confused and suspect the command has failed silently if they see no output.

cmd/dbc/list.go
zeroshade: I was trying to avoid allowing `--level env` here, the env level will show up if you don't specify a particular level, but it's a fair point to allow using it here to only show that level. I suppose we can still disallow it for install though

cmd/dbc/list.go
zeroshade: I agree with @ianmcook that we want *something* and it should go to stderr to allow `wc -l`

zeroshade:
> I'm -1 on this. I think the current output is good, but I can see a case for maybe adding a --quiet, -q option that prints the output like you suggest (just the slugs, one per line). But if we add that, we should add it to the search subcommand too I think.

We could invert it and have the default be the simple list and add a `-v/--verbose` flag? Or I wouldn't use `-q` for it, maybe `-s/--simple`? or some other flag?

amoeba:
I think we can skip having any sort of simpler variant of the list command for now, users can filter with a pipe or jq.

cmd/dbc/list.go
amoeba: Maybe the best thing to do is to only allow user/system (like it is now) but say environment drivers will be included when set.

ianmcook:
> > I'm -1 on this. I think the current output is good, but I can see a case for maybe adding a --quiet, -q option that prints the output like you suggest (just the slugs, one per line). But if we add that, we should add it to the search subcommand too I think.
> 
> We could invert it and have the default be the simple list and add a `-v/--verbose` flag? Or I wouldn't use `-q` for it, maybe `-s/--simple`? or some other flag?

I'm -1 on inverting it. The style of output that this currently makes isn't verbose enough for the `--verbose` flag to be right. It would feel inconsistent with what `dbc search --verbose` prints.

I think it's OK to defer this until later, and consider it in the scope of other subcommands too, not just `list`.

zeroshade:
the failure here is likely due to the active snapcraft outage going on (status.snapcraft.io)

zeroshade:
@ianmcook @amoeba are we good here then?

amoeba (approved):
+1 as long as we're okay with the docs being ahead of the release by a bit

ianmcook:
Are the relevant bits from https://github.com/columnar-tech/dbc/pull/334 all reverted here @zeroshade ?

zeroshade:
Yea, the "list" -> "search" subcommand suggestion was removed, though the infrastructure for it is still there so we can add more suggestions in the future

ianmcook (approved):

PR #375, opened 2026-05-05
fix(fslock): delete lock file on Release

zeroshade:
## Summary
- `Release` now deletes the lock file instead of leaving it on disk.
- Closes the TOCTOU that previously motivated keeping the file: a naive delete-on-release races with a concurrent `Acquire` that has opened the path but not yet taken the lock, leaving two processes with locks on different inodes at the same path.
- Unix: unlink-before-close so new openers can't see our inode, plus an inode recheck after `flock` — if the path no longer points at our fd's inode, we release, close, and reopen via a retry loop.
- Windows: `Close` then `os.Remove`. Go opens files without `FILE_SHARE_DELETE`, so `Remove` only succeeds once no other process holds a handle — we swallow the error if cleanup can't happen this round.

## Test plan
- [x] `go test ./internal/fslock/ -race -count=1 -v` — all pass, including two new tests:
  - `TestReleaseRemovesFile` verifies the file is gone after `Release` (would have failed on old code).
  - `TestConcurrentAcquireIsExclusive` spawns 8 goroutines that acquire/hold/release and asserts the max concurrent holder count is 1, exercising the delete/reopen path.
- [x] `GOOS=windows go build ./internal/fslock/` compiles cleanly.
- [x] Full `go test ./... -short` passes.

### DIFF
--- a/internal/fslock/fslock.go
+++ b/internal/fslock/fslock.go
@@ -20,11 +20,6 @@ import "os"
 
 // Lock represents an acquired advisory file lock.
 type Lock struct {
-	f *os.File
-}
-
-// Release releases the lock and closes the file. The lock file is left on
-// disk so that concurrent waiters do not race on a deleted path.
-func (l Lock) Release() error {
-	return l.f.Close()
+	f    *os.File
+	path string
 }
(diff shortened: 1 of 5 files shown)


amoeba (approved):
Tested manually and it looks good.

PR #377, opened 2026-05-08
docs: add `dbc list` validation step to driver install guides

dataders:
## Summary

Adds a `dbc list` step to the two driver-install walkthroughs so readers can confirm the driver landed where they expect, right after running `dbc install`.

- `docs/getting_started/first_steps.md` — after the BigQuery install
- `docs/guides/installing.md` — after the MySQL install

`dbc list` shipped in [v0.3.0](https://columnar.tech/blog/announcing-dbc-0.3.0/) and is documented at [reference/cli.md#list](https://docs.columnar.tech/dbc/reference/cli/#list); these guides predate it and don't yet mention it.

## Note on `since_version`

I did **not** add a `{{ since_version('v0.3.0') }}` marker next to the new `dbc list` snippets. The reference page uses one, but these guides read as "current behavior" and don't mark other recent additions either. Happy to add the marker if maintainers prefer to flag it for readers on older versions.

## Test plan

- [ ] `mkdocs serve` and confirm both pages render the new code blocks correctly
- [ ] Verify the `../reference/cli.md#list` anchor links resolve

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/getting_started/first_steps.md
+++ b/docs/getting_started/first_steps.md
@@ -104,6 +104,14 @@ Installed bigquery 1.0.0 to /Users/user/Library/Application Support/ADBC/Drivers
 
 The BigQuery ADBC driver is now installed and usable by any driver manager.
 
+To confirm the driver is installed, run [`dbc list`](../reference/cli.md#list):
+
+```console
+$ dbc list
+DRIVER    VERSION  LEVEL  LOCATION
+bigquery  1.0.0    user   /Users/user/Library/Application Support/ADBC/Drivers
+```
+
 For more information on on how to find drivers, see the [Finding Drivers](../guides/finding_drivers.md) guide.
 
 ## Installing a Driver Manager
--- a/docs/getting_started/first_steps.md
+++ b/docs/getting_started/first_steps.md
@@ -104,14 +104,6 @@ Installed bigquery 1.0.0 to /Users/user/Library/Application Support/ADBC/Drivers
 
 The BigQuery ADBC driver is now installed and usable by any driver manager.
 
-To confirm the driver is installed, run [`dbc list`](../reference/cli.md#list):
-
-```console
-$ dbc list
-DRIVER    VERSION  LEVEL  LOCATION
-bigquery  1.0.0    user   /Users/user/Library/Application Support/ADBC/Drivers
-```
-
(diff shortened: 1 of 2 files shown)


amoeba (commented):
Thanks for the PR @dataders.

I'm on the fence about whether we want to train users to run `dbc list` after install but we're building our docs for people and agents so having this may make sense. Most of the package managers I use just tell users to run `install` and then the user builds a mental model that their package was installed without needing to run a separate command (i.e., guides don't say "Run pip list" after they tell you to "Run pip install").

Is this a documentation gap you ran into yourself or is the PR just for completeness?

dataders:
@amoeba yeah I understand your hesitance. I did double check `pip` and `uv` docs and you're right: none of the initial guides mention `list` or `show` for verification.

Maybe I just still have PTSD from my previous experience (https://github.com/columnar-tech/dbc/pull/332) of trying to figure out how to install an older version of the duckdb driver.

maybe there's a space for a new "auditing installed drivers" or a "troubleshooting" section. [The agent skill](https://github.com/columnar-tech/skills/blob/main/skills/adbc/SKILL.md) doesn't have any reference to auditing what is installed

docs/getting_started/first_steps.md
```diff
@@ -104,6 +104,14 @@ Installed bigquery 1.0.0 to /Users/user/Library/Application Support/ADBC/Drivers
...
 The BigQuery ADBC driver is now installed and usable by any driver manager.
 
+To confirm the driver is installed, run [`dbc list`](../reference/cli.md#list):
+
+```console
+$ dbc list
+DRIVER    VERSION  LEVEL  LOCATION
+bigquery  1.0.0    user   /Users/user/Library/Application Support/ADBC/Drivers
+```
+
```
amoeba:
```suggestion
```

Maybe removing this and keeping the one in installing is a good compromise?

amoeba (commented):
An auditing/troubleshooting section is a good idea. Re: the agent skills, we have an issue for adding dbc list to it: https://github.com/columnar-tech/skills/issues/20.

What do you think about my suggestion here?

dataders:
> What do you think about my suggestion here?

sounds good to me! I've got no strong opinions here, I defer to you. Thanks for hearing me out

amoeba (approved):
Thanks for taking the time to file a PR!

PR #379, opened 2026-05-09
ci(release): enable winget release automation

eitsupi:
Since creating PR to the winget-pkgs repo is not automated, someone needs to do it.

Also, winget doesn't currently distinguish between pre-release and newer versions, so it might be better to skip creating PRs for pre-release versions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -231,6 +231,7 @@ jobs:
           MACOS_NOTARY_KEY: ${{ secrets.MACOS_NOTARY_KEY }}
           MACOS_NOTARY_KEY_ID: ${{ secrets.MACOS_NOTARY_KEY_ID }}
           DBC_TAP_GITHUB_PAT: ${{ secrets.DBC_TAP_GITHUB_PAT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -231,6 +231,7 @@ jobs:
           MACOS_NOTARY_KEY: ${{ secrets.MACOS_NOTARY_KEY }}
           MACOS_NOTARY_KEY_ID: ${{ secrets.MACOS_NOTARY_KEY_ID }}
           DBC_TAP_GITHUB_PAT: ${{ secrets.DBC_TAP_GITHUB_PAT }}
+          DBC_WINGET_GITHUB_PAT: ${{ secrets.DBC_WINGET_GITHUB_PAT }}
```
eitsupi, line 234: We need to set a token here that has write permissions to https://github.com/columnar-tech/winget-pkgs.

zeroshade: Added a token and secret for this. Should be good now

eitsupi: Thanks!

amoeba (approved):
This looks good to me, thanks @eitsupi.

PR #381, opened 2026-05-18
feat: publish dbc to npm

### LINKED ISSUE #355: Distribute dbc on npm
This would enable users to install dbc with

```sh
npm install -g @columnar-tech/dbc
```

or to install drivers with

```sh
npx @columnar-tech/dbc install <driver>
```

This would be a great complement to the new [ADBC driver manager on npm](https://www.npmjs.com/package/@apache-arrow/adbc-driver-manager).

amoeba:
Adds support for publishing dbc to npm by,

1. Adding a create_packages.js script and shell packages that can create npm packages for each platform
2. Integrating this into the release.yml workflow so we publish to npm automatically as part of a release

There are two options for shipping dbc on npm, 

1. A single package with a postinstall script that detects the user's platform and downloads and installs dbc at install time
4. One package for each platform, with a wrapper package for the user to install

Option 1 is the only one supported by goreleaser but option 2 is better IMO because we ship the dbc binary in the npm package. This means installation isn't making an extra request to a separate origin and it also means the package can be installed with scripts disabled.

Closes https://github.com/columnar-tech/dbc/issues/355


.github/workflows/release.yml
```diff
@@ -250,3 +249,47 @@ jobs:
             --value="${RELEASE_VERSION}"
         env:
           RELEASE_VERSION: ${{ fromJSON(steps.gorelease.outputs.metadata).version }}
+
+  publish_npm:
```
zeroshade, line 279: could we just use https://goreleaser.com/customization/publish/npm/ instead?

amoeba:
I changed how the npm artifacts are built so as little as possible is checked into the repo. create_package.js now creates the entire file structure, including package.json. This is ready for another review.

zeroshade (approved):
LGTM, my only question is whether there is any ability to add a "snapshot" style dry-run path to the CI somehow (for example, we have --snapshot for goreleaser so that we can ensure via CI that we don't break the release process).

amoeba:
I added a step that just creates the npm packages to the `snapshot` job in https://github.com/columnar-tech/dbc/pull/381/commits/bcec6d5fde0a770b54810dc1c9b45b545ceb42da.

amoeba:
I'm going to test the update to the snapshot job in a separate draft pr before merging just to make sure it works.

amoeba:
Okay, the snapshot job now has a test that actually installs the wrapper package and makes sure the equivalent of `dbc --version` runs successfully. I'll merge this.

PR #383, opened 2026-05-19
fix: try to enable vt or disable colors on startup on Windows

amoeba:
Fixes an issue a user reported privately where they were seeing ANSI escapes in dbc's output. I was able to reproduce their issue by setting `HKCU\Console\VirtualTerminalLevel` to `1` and I came up with this as a fix. When dbc starts on Windows, try to enable VT processing if it's not on and, if we fail, set `NO_COLOR` so the user doesn't get garbled output. I wasn't able to figure out how an admin could disable VT processing so that branch wasn't tested.

### DIFF
--- a/cmd/dbc/main.go
+++ b/cmd/dbc/main.go
@@ -238,6 +238,9 @@ func main() {
 		args cmds
 	)
 
+	// Attempt to enable VT processing or disable colors if we can't
+	initWindowsConsole()
+
 	if err := initDBCClient(); err != nil {
 		fmt.Fprintf(os.Stderr, "error initializing client: %v\n", err)
 		os.Exit(1)
--- a/cmd/dbc/main.go
+++ b/cmd/dbc/main.go
@@ -239,7 +239,7 @@ func main() {
 	)
 
 	// Attempt to enable VT processing or disable colors if we can't
-	initWindowsConsole()
+	initVTProcessing()
 
 	if err := initDBCClient(); err != nil {
 		fmt.Fprintf(os.Stderr, "error initializing client: %v\n", err)
--- a/cmd/dbc/terminal_windows.go
+++ b/cmd/dbc/terminal_windows.go
@@ -1,5 +1,3 @@
-//go:build windows
-
 // Copyright 2026 Columnar Technologies Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build windows
+
 package main
 
 import (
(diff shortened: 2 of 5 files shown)


cmd/dbc/terminal_windows.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Get current console mode
+	err := windows.GetConsoleMode(handle, &mode)
+	if err != nil {
+		return false
+	}
+
+	// Check if VT processing is already enabled
+	if mode&ENABLE_VIRTUAL_TERMINAL_PROCESSING != 0 {
+		return true
```
amoeba, line 50: On a normal system, we always hit this return.

cmd/dbc/main.go
```diff
@@ -238,6 +238,9 @@ func main() {
 		args cmds
 	)
 
+	// Attempt to enable VT processing or disable colors if we can't
+	initWindowsConsole()
```
zeroshade: since we're differentiating this between systems using `go:build` can we pick a better name for this? like `initVTProcessing` or something like that?

cmd/dbc/windows_console.go
```diff
@@ -0,0 +1,72 @@
+//go:build windows
```
zeroshade: this goes below the copyright header but above the `package` line. It just needs to have empty lines surrounding it.

Alternatively, if the filename just ends with `_windows.go` then you don't even actually need this line 😄

cmd/dbc/windows_console.go
```diff
@@ -0,0 +1,72 @@
...
+func initWindowsConsole() {
+	if !enableVTProcessing(os.Stdout) {
+		disableColors()
+		return
+	}
+
+	if !enableVTProcessing(os.Stderr) {
+		disableColors()
+		return
+	}
```
zeroshade:
```suggestion
	if !enableVTProcessing(os.Stdout) || !enableVTProcessing(os.Stderr) {
		disableColors()
		return
	}
```

cmd/dbc/windows_console_unix.go
```diff
@@ -0,0 +1,19 @@
+//go:build !windows
```
zeroshade: move this down below the copyright header but above the package line. it just needs to be surrounded by empty lines. See https://github.com/columnar-tech/dbc/blob/1f119d0488c34b9db1b909c954aa3c834c02b588/config/dirs_unixlike.go#L15 for example

zeroshade (changes requested):

zeroshade:
Just a few nitpicks :smile:

amoeba:
I reorganized things in 0c9b0cb2b2ac1906c28a421222e16a93fbf0e1a7. How does that look?

zeroshade (approved):

zeroshade:
LGTM

PR #390, opened 2026-05-29
test(install): cover downgrade by reinstalling an older version

zeroshade:
## Summary

Adds `TestReinstallDowngradeVersion` to `cmd/dbc/install_test.go`, mirroring the existing `TestReinstallUpdateVersion` but exercising the opposite direction: install latest, then reinstall pinned to an older version.

Before this change, no test verified that downgrading a driver:

1. Removes the previously-installed newer version (including its install directory)
2. Lays down the older version's files on disk
3. Leaves the driver discoverable via the manifest

The conflict-and-replace path (`Removed conflicting driver: …`) was only covered for upgrades (1.0.0 → 1.1.0). This adds symmetric coverage for downgrades (1.1.0 → 1.0.0), exercising the same code path with the version comparison reversed.

## Test plan

```
$ go test -run 'TestSubcommandsEnv/TestReinstall(Update|Downgrade)Version' -v ./cmd/dbc/...
=== RUN   TestSubcommandsEnv/TestReinstallDowngradeVersion
=== RUN   TestSubcommandsEnv/TestReinstallUpdateVersion
--- PASS: TestSubcommandsEnv (0.01s)
    --- PASS: TestSubcommandsEnv/TestReinstallDowngradeVersion (0.01s)
    --- PASS: TestSubcommandsEnv/TestReinstallUpdateVersion (0.00s)
PASS
```

`go vet ./cmd/dbc/...` is clean.

## Notes

The expected install directory for v1.0.0 is `test-driver-1/` (not `test-driver-1.0/`) because the directory name is derived from the tarball filename (`test-driver-1.tar.gz` for v1.0.0, `test-driver-1.1.tar.gz` for v1.1.0) per `config.InstallDriver`. No fixture changes are needed.

### DIFF
--- a/cmd/dbc/install_test.go
+++ b/cmd/dbc/install_test.go
@@ -99,20 +99,20 @@ func (suite *SubcommandTestSuite) TestReinstallUpdateVersion() {
 }
 
 func (suite *SubcommandTestSuite) TestReinstallDowngradeVersion() {
-	m := InstallCmd{Driver: "test-driver-1"}.
+	m := InstallCmd{Driver: "test-driver-1", Level: suite.configLevel}.
 		GetModelCustom(testBaseModel())
 	suite.validateOutput("\r[✓] searching\r\n[✓] downloading\r\n[✓] installing\r\n[✓] verifying signature\r\n",
-		"\nInstalled test-driver-1 1.1.0 to "+suite.tempdir, suite.runCmd(m))
+		"\nInstalled test-driver-1 1.1.0 to "+suite.Dir(), suite.runCmd(m))
 	suite.driverIsInstalledWithVersion("test-driver-1", "1.1.0", true)
 
-	m = InstallCmd{Driver: "test-driver-1<=1.0.0"}.
+	m = InstallCmd{Driver: "test-driver-1<=1.0.0", Level: suite.configLevel}.
 		GetModelCustom(testBaseModel())
 	suite.validateOutput("\r[✓] searching\r\n[✓] downloading\r\n[✓] installing\r\n[✓] verifying signature\r\n",
-		"\nRemoved conflicting driver: test-driver-1 (version: 1.1.0)\nInstalled test-driver-1 1.0.0 to "+suite.tempdir,
+		"\nRemoved conflicting driver: test-driver-1 (version: 1.1.0)\nInstalled test-driver-1 1.0.0 to "+suite.Dir(),
 		suite.runCmd(m))
 
 	suite.Equal([]string{"test-driver-1/test-driver-1-not-valid.so",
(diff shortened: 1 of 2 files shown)


cmd/dbc/install_test.go
```diff
@@ -98,6 +98,22 @@ func (suite *SubcommandTestSuite) TestReinstallUpdateVersion() {
...
+		"\nInstalled test-driver-1 1.1.0 to "+suite.tempdir, suite.runCmd(m))
+
+	m = InstallCmd{Driver: "test-driver-1<=1.0.0"}.
+		GetModelCustom(testBaseModel())
+	suite.validateOutput("\r[✓] searching\r\n[✓] downloading\r\n[✓] installing\r\n[✓] verifying signature\r\n",
+		"\nRemoved conflicting driver: test-driver-1 (version: 1.1.0)\nInstalled test-driver-1 1.0.0 to "+suite.tempdir,
+		suite.runCmd(m))
+
+	suite.Equal([]string{"test-driver-1/test-driver-1-not-valid.so",
+		"test-driver-1/test-driver-1-not-valid.so.sig", "test-driver-1.toml"}, suite.getFilesInTempDir())
```
ianmcook: Should the test also look inside the manifest to confirm that the driver version number is <= 1.0.0?

ianmcook: Currently the only way this actually validates that the driver is replaced is by validating the dbc output, but hypothetically the output could be wrong.

amoeba: I added 6cd95121d150f1ccb99451cfa972ea3614f007bc to address this. Feel free to rework it @zeroshade.

zeroshade: Updated and got the tests passing.

zeroshade: Feel free to take another look @amoeba

cmd/dbc/install_test.go
```diff
@@ -98,6 +98,28 @@ func (suite *SubcommandTestSuite) TestReinstallUpdateVersion() {
 		"test-driver-1.1/test-driver-1-not-valid.so.sig", "test-driver-1.toml"}, suite.getFilesInTempDir())
 }
 
+func (suite *SubcommandTestSuite) TestReinstallDowngradeVersion() {
+	if runtime.GOOS == "windows" && (suite.configLevel == config.ConfigUser || suite.configLevel == config.ConfigSystem) {
+		suite.T().Skip("Driver manifest is stored in the registry on Windows for User and System config levels, so the file-list assertion does not apply")
```
ianmcook: Obvious next question: can we check the registry on Windows instead of just skipping the test? :)

amoeba: Fixed in https://github.com/columnar-tech/dbc/pull/390/changes/77e7a3402411d9c6b3cf0a635b6386859bb08396.

amoeba (approved):

PR #392, opened 2026-06-10
docs: add new drivers and languages

ianmcook:
This catches up the dbc docs with what's recently been added to [columnar.tech/dbc](https://columnar.tech/dbc).


amoeba (approved):
Thanks for doing this.

PR #399, opened 2026-06-30
fix(wasm): driver discovery on Windows hosts

### LINKED ISSUE #396: WASM: complete Windows-host support for @columnar-tech/dbc-wasm
Tracks finishing Windows-host support for the `@columnar-tech/dbc-wasm` package (introduced in #395). Linux and macOS are validated in CI; Windows is experimental and **dropped from the WASM CI matrix** until the issue below is resolved.

## Fixed on the way (already on the branch / in #395)
- **Env-size overflow:** `boot.cjs` no longer forwards the full `process.env` into the wasm runtime. A GitHub Actions Windows env is large enough to overflow `wasm_exec.js`'s combined argv+env size cap, making `go.run()` throw before startup. Now only the vars `GOOS=js` needs are forwarded (`HOME`/`XDG_*`/`TMPDIR`), via `curateGoEnv()`.
- **Temp dir:** `GOOS=js` `os.TempDir()` reads only `$TMPDIR` (else `/tmp`, missing on Windows). `curateGoEnv()` maps `TMPDIR` from `%TMPDIR%||%TEMP%||%TMP%` and normalizes backslashes.

With those, the Windows runtime boots and `install()` completes.

## Remaining blocker
`listInstalled()` returns `[]` on Windows even though `install()` wrote the manifest. Repro (from the WASM windows-latest smoke):

- `installDir` contains `test-driver-1.toml` at the top level (confirmed via Node `fs.readdirSync`).
- `dbc.listInstalled(installDir)` → `config.FindDriverConfigsIn(\"C:/.../installDir\")` → `loadDir` → returns no drivers.

Switching `loadDir` from `fs.Glob(os.DirFS(dir), \"*.toml\")` to `os.ReadDir(dir)` did **not** help (identical empty result), so it's not the `os.DirFS` abstraction. Leading hypothesis: under `GOOS=js`, a normalized drive-letter path like `C:/Users/...` is not treated as absolute (`filepath.IsAbs` is false on the unix-like js target), so the runtime's path resolution for discovery diverges from where `install` actually wrote. Needs Go-side, syscall-level diagnostics on a real Windows runner (os.Stat/ReadDir/Open results + the exact resolved path inside `loadDir`).

## Action
- [ ] Re-add `windows-latest` to `.github/workflows/wasm.yml` once discovery works.
- [ ] Pin down the `GOOS=js` + Windows drive-path resolution behavior in `config` driver discovery.
- [ ] Decide whether the loader should canonicalize Windows locations to a wasm-fs-absolute form before handing them to Go.

zeroshade:
## Summary

- Fix `listInstalled` returning empty on Windows hosts when running via WASM. The root cause is Go's js/wasm `syscall.Open` rejecting `O_DIRECTORY` (Node.js doesn't expose `fs.constants.O_DIRECTORY` on Windows), which breaks `os.ReadDir`/`os.DirFS`/`fs.Glob`.
- Split `loadDir` into build-tagged files: `loaddir_js.go` uses `os.Open` + `File.ReadDir` (avoids `O_DIRECTORY`); `loaddir_other.go` keeps the original `os.DirFS` + `fs.Glob` path.
- Re-enable `windows-latest` in the WASM CI workflow.

Closes #396

## Test plan

- [x] `node packages/npm-wasm/test/smoke.cjs` passes on Windows (install + listInstalled + uninstall round-trip)
- [x] `node packages/npm-wasm/test/worker.test.cjs` passes on Windows
- [x] `go test ./config/...` passes natively on Windows
- [x] `go vet ./config/...` passes for both `GOOS=js` and native targets
- [x] CI passes on all three platforms (ubuntu, macos, windows)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -55,10 +55,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # windows-latest is intentionally excluded: the WASM Windows-host path is
-        # experimental and not yet passing (driver discovery under GOOS=js — see
-        # https://github.com/columnar-tech/dbc/issues/396). Re-add when fixed.
-        os: [ 'ubuntu-latest', 'macos-latest' ]
+        os: [ 'ubuntu-latest', 'macos-latest', 'windows-latest' ]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/config/config.go
+++ b/config/config.go
@@ -148,28 +148,6 @@ func EnsureLocation(cfg Config) (string, error) {
 	return loc, nil
 }
 
-func loadDir(dir string) (map[string]DriverInfo, error) {
-	if _, err := os.Stat(dir); err != nil {
-		return nil, err
-	}
-
-	ret := make(map[string]DriverInfo)
-
-	fsys := os.DirFS(dir)
-	matches, _ := fs.Glob(fsys, "*.toml")
-	for _, m := range matches {
-		p := filepath.Join(dir, m)
-		di, err := loadDriverFromManifest(filepath.Dir(p), filepath.Base(p))
-		if err != nil {
-			continue
-		}
-
-		di.FilePath = filepath.Dir(p)
-		ret[di.ID] = di
-	}
-	return ret, nil
-}
-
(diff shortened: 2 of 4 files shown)


amoeba (approved):
This seems fine. The two variants share enough code that we could extract just the platform-variable part but this is simple and it works.

PR #435, opened 2026-07-22
chore: update fallback driver docs URLs

ianmcook:


ianmcook:
Do we still need these hard-coded? Should we add the docs URLs for all the other known drivers?

amoeba:
We don't need this functionality any more. The CDN now validates on commit and in CI that every driver in the CDN has a documentation URL. I'll merge this and pull the fallback functionality out.

amoeba (approved):

PR #447, opened 2026-07-27
fix: improve error message for locking failures

### LINKED ISSUE #446: errors about lockfiles in dbc install/uninstall when not running with sudo/admin
When I try to uninstall a driver in an unprivileged PowerShell shell. I get an error about access being denied to a lock file:

```
PS > dbc uninstall --level system foo
  Error: another dbc operation is in progress: fslock: open C:\Program Files\ADBC\Drivers\.dbc.install.lock: open C:\Program
  Files\ADBC\Drivers\.dbc.install.lock: Access is denied.
```

I think this error would be misleading and we should instead err about a permissions issue instead. May be related to https://github.com/columnar-tech/dbc/issues/445.

This also happens on macOS so I assume Linux as well.

Environment:
- dbc 0.3.0
- Windows 11, macOS

amoeba:
Certain dbc subcommands use filesystem locks to avoid problems related to running multiple copies of dbc at the same time. The locking mechanism currently doesn't distinguish between a real locking failure and not being able to create the lock file in the first place. This can happen if the user tries to run a command like `dbc install --level system foo` without admin privs.

This PR adds a specialized error `ErrLockContended` for lock contention that our fslock implementation can use as a sentinel error and wraps existing calls in a helper function `acquireLock` that is responsible for handling errors from fslock and printing something more useful to the user than what's currently printed. Additionally, the new, more helpful error also suggests trying the command again with the appropriate mechanism for the user's operating system.

Closes https://github.com/columnar-tech/dbc/issues/446

### DIFF
--- a/cmd/dbc/add.go
+++ b/cmd/dbc/add.go
@@ -27,7 +27,6 @@ import (
 	"charm.land/lipgloss/v2"
 	"github.com/Masterminds/semver/v3"
 	"github.com/columnar-tech/dbc/config"
-	"github.com/columnar-tech/dbc/internal/fslock"
 	"github.com/columnar-tech/dbc/internal/jsonschema"
 	"github.com/pelletier/go-toml/v2"
 )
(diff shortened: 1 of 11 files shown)


zeroshade (approved):
lgtm

zeroshade (approved):
lgtm

amoeba:
Thanks!

PR #451, opened 2026-07-28
chore: add winget steps to release checklist

amoeba:
Updates the release checklist to separate it into pre- and post- sections and adds our steps for updating winget.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -77,7 +77,7 @@ Note: If/when we start publishing dbc binaries to WinGet for arm64, we'll probab
 
 ```sh
 komac update \
-      --version ${VERSION}
+      --version ${VERSION} \
       --urls https://github.com/columnar-tech/dbc/releases/download/v${VERSION}/dbc-${VERSION}-x64.msi \
       -- \
       Columnar.dbc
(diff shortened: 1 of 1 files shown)


RELEASE.md
eitsupi, line 1: IIUC, we shouldn't need to use Komac from our local machine anymore.
If #379 works correctly, a new branch will be created at https://github.com/columnar-tech/winget-pkgs, and we can then open a pull request upstream from that branch.

amoeba: Oh, right! I'll update the checklist. Thanks for reminding me.

eitsupi: (I believe this setting has never used before, so it might fail the first time and we should run Komac for recovery.)

amoeba: How's cb082aa3c46bfb970997c2d901601ef4c1b8be54?

eitsupi: Looks good!

amoeba: > (I believe this setting has never used before, so it might fail the first time and we should run Komac for recovery.)

k, noted.

eitsupi (approved):
Thanks!

Issue #50, opened 2025-09-18
Idea: Add options to install drivers for other platforms


amoeba:
Right now when you install a driver, dbc looks at your GOOS and GOOARCH to determine what driver to go look for. It might be useful if users can install drivers for another combo, such as running dbc on windows-arm and installing windows-amd64 drivers (for eventual running under windows-amd64 programs)

wbeardall:
Is this still under consideration? The feature would be very useful for our build pipelines!

zeroshade:
It's definitely on our roadmap, but I do not currently have a timeline for when we will be developing it. We definitely accept contributions though!

wbeardall:
Hopefully the above is helpful :)
Have also added a follow-up PR for `dbc list`, addressing potential confusion due to aliasing host- and cross-platform drivers in the printed summary [#467](https://github.com/columnar-tech/dbc/pull/467)

zeroshade:
sorry for the delay here, been busy with a bunch of other things. Gonna take a look and review these now. Thanks for your patience!

wbeardall:
No worries! Thanks for considering the contribution :)

PR #466, opened 2026-08-22
feat: added optional platform selection for driver install

wbeardall:
## Summary

This is a contribution in response to [Issue #50](https://github.com/columnar-tech/dbc/issues/50)

The idea is to expand platform resolution, enabling users to install drivers which target non-host platforms when using the `dbc install` command.

## Goals and Constraints

- Do not change default `dbc install` behaviour; non-host always opt-in
- WASM lib invocations are unmodified; the WASM / JS interface does not expose platform selection, as this PR is only intended to target 
- Non-host targets are validated ASAP during CLI parse; this does mean that the set of accepted platforms is hard-coded in `platform.go`, worth considering in case someone ends up adding weird and wonderful supported platforms to the CDN
- Selected platforms are *not* validated against the available set of platforms for the specific driver requested (e.g. someone requests `windows_arm64` for `sqlite`. In this case, the standard `Error: no package found for platform 'windows_arm64'` path occurs.

## Potential Concerns and Considerations

The main concern I have currently is that `dbc list` reads installed manifests without confirming that the driver is actually installed *for the host machine*. This has the potential for user confusion, as they might have a non-host driver installed, and assume that the entry in `dbc list` means that they can use that driver with ADBC now!

I deliberately haven't changed the `dbc list` behaviour, because it has the potential for a bigger direct impact on user experience than the rest of this PR. That said, this is how I'd go about modifying `dbc list`:

1. Add a `--all-platforms` flag to the `dbc list command`
2. In absence of this flag, `dbc list` should filter out any drivers without a host lib explicitly linked in the `<driver>.toml`, and then print exactly as current
3. With the `--all-platforms` flag, the `dbc list` table would expand with another `PLATFORM` column, with the host platform being marked explicitly (e.g. `linux_amd64 (*)`)

I've put together a follow-up [PR #467](https://github.com/columnar-tech/dbc/pull/467) implementing this. 

### `config/platform_test.go` (new file)

TestPlatformUnmarshalText
  - All six valid tuples unmarshal successfully
  - Invalid tuples fail with "unknown platform" and list valid values
TestPlatformResolve
  - Empty Platform resolves to PlatformTuple() (host)
  - Explicit linux_amd64 resolves to itself
 
### `config/config_api_test.go`

TestInstallDriver/records_explicit_platform_in_manifest
  - InstallDriver with linux_amd64 keys Driver.shared under that tuple
  - Shared lib file exists on disk
  - Host platform key is empty when it differs from requested platform

### `cmd/dbc/main_test.go`

TestInstallInvalidPlatformRejectedAtParse
  - dbc install --platform noos_noarch mysql fails at argv parse
  - Error contains "unknown platform" and "valid values are:"
TestInstallHelpMentionsVersionConstraints (extended)
  - Help output includes --platform

### `cmd/dbc/install_test.go`

TestInstallWithPlatform (SubcommandTestSuite)
  - End-to-end CLI install with Platform: linux_amd64
  - Success output, manifest keyed under linux_amd64, file on disk
  - Host platform key absent when it differs from requested platform

### `client_methods_test.go`

TestClientInstall/installs_driver_for_explicit_platform
  - Client.Install(..., config.Platform("linux_amd64")) succeeds
  - Shared path recorded under requested platform key

### Updated Tests

TestClientInstall, TestClientUninstall, TestInstallDriver/success,
TestInstallDriver/invalid_tarball — updated for new Install/InstallDriver signatures

## Notes

- Code style: I couldn't see any specific precommit / lint-staged / etc. configurations, so I've not `gofmt`-ed the PR specifically!
- Tests all pass on my machine :)

### DIFF
--- a/client_methods_test.go
+++ b/client_methods_test.go
@@ -119,15 +119,25 @@ func TestClientInstall(t *testing.T) {
 	}
 
 	t.Run("installs driver successfully", func(t *testing.T) {
-		manifest, err := c.Install(t.Context(), cfg, "test-driver-1")
+		manifest, err := c.Install(t.Context(), cfg, "test-driver-1", "")
 		require.NoError(t, err)
 		require.NotNil(t, manifest)
 		assert.Equal(t, "test-driver-1", manifest.DriverInfo.ID)
 		assert.NotNil(t, manifest.DriverInfo.Version)
 	})
 
+	t.Run("installs driver for explicit platform", func(t *testing.T) {
+		const platform = "linux_amd64"
+		manifest, err := c.Install(t.Context(), cfg, "test-driver-1", config.Platform(platform))
+		require.NoError(t, err)
+		require.NotNil(t, manifest)
+		sharedPath := manifest.DriverInfo.Driver.Shared.Get(platform)
+		assert.NotEmpty(t, sharedPath)
+		assert.FileExists(t, sharedPath)
+	})
+
 	t.Run("returns error for nonexistent driver", func(t *testing.T) {
-		_, err := c.Install(t.Context(), cfg, "nonexistent-driver")
+		_, err := c.Install(t.Context(), cfg, "nonexistent-driver", "")
 		assert.Error(t, err)
 		assert.Contains(t, err.Error(), "nonexistent-driver")
 	})
@@ -144,7 +154,7 @@ func TestClientUninstall(t *testing.T) {
 			Location: tmpDir,
 		}
 
-		_, err := c.Install(t.Context(), cfg, "test-driver-1")
+		_, err := c.Install(t.Context(), cfg, "test-driver-1", "")
 		require.NoError(t, err)
 
 		manifestPath := filepath.Join(tmpDir, "test-driver-1.toml")
--- a/cmd/dbc/install.go
+++ b/cmd/dbc/install.go
@@ -66,6 +66,7 @@ type InstallCmd struct {
 	// URI    url.URL `arg:"-u" placeholder:"URL" help:"Base URL for fetching drivers"`
 	Driver             string             `arg:"positional,required" help:"Driver to install, optionally with a version constraint (for example: mysql, mysql=0.1.0, mysql>=1,<2)"`
 	Level              config.ConfigLevel `arg:"-l" help:"Config level to install to (user, system)"`
+	Platform           config.Platform    `arg:"--platform" placeholder:"TUPLE" help:"Platform tuple to install for (e.g. linux_amd64). Defaults to the host platform."`
 	Json               bool               `arg:"--json" help:"Print output as JSON instead of plaintext"`
 	JsonStreamProgress bool               `arg:"--json-stream-progress" help:"Stream progress events as JSON lines (implies --json)"`
 	NoVerify           bool               `arg:"--no-verify" help:"Allow installation of drivers without a signature file"`
@@ -89,6 +90,7 @@ func (c InstallCmd) GetModelCustom(baseModel baseModel) tea.Model {
 	}
 	return progressiveInstallModel{
 		Driver:             c.Driver,
+		platform:           c.Platform.Resolve(),
 		NoVerify:           c.NoVerify,
 		jsonOutput:         c.Json || c.JsonStreamProgress,
 		jsonStreamProgress: c.JsonStreamProgress,
@@ -111,12 +113,12 @@ func (c InstallCmd) GetModel() tea.Model {
 	return c.GetModelCustom(defaultBaseModel())
 }
 
-func verifySignature(m config.Manifest, noVerify bool) error {
+func verifySignature(m config.Manifest, platform string, noVerify bool) error {
 	if m.Files.Driver == "" || noVerify {
 		return nil
 	}
 
-	path := filepath.Dir(m.Driver.Shared.Get(config.PlatformTuple()))
+	path := filepath.Dir(m.Driver.Shared.Get(platform))
 
 	lib, err := os.Open(filepath.Join(path, m.Files.Driver))
 	if err != nil {
@@ -215,6 +217,7 @@ type progressiveInstallModel struct {
 	baseModel
 
 	Driver             string
+	platform           string
 	VersionInput       *semver.Version
 	NoVerify           bool
 	jsonOutput         bool
@@ -353,8 +356,8 @@ func (m progressiveInstallModel) FinalOutput() string {
 			installStatus.Message = m.postInstallMessage
 		}
 
-		if !m.insecureNoChecksum && m.installedDriverInfo.Driver.Shared.Get(config.PlatformTuple()) != "" {
-			driverPath := m.installedDriverInfo.Driver.Shared.Get(config.PlatformTuple())
+		if !m.insecureNoChecksum && m.installedDriverInfo.Driver.Shared.Get(m.platform) != "" {
+			driverPath := m.installedDriverInfo.Driver.Shared.Get(m.platform)
 			chksum, err := checksum(driverPath)
 			if err != nil && m.jsonOutput {
 				return marshalEnvelope("error", jsonschema.ErrorResponse{
@@ -415,14 +418,14 @@ func (m progressiveInstallModel) searchForDriver(list []dbc.Driver) (tea.Model, 
 	return m, func() tea.Msg {
 		if vers != nil {
 			vers.IncludePrerelease = m.Pre
-			pkg, err := d.GetWithConstraint(vers, config.PlatformTuple())
+			pkg, err := d.GetWithConstraint(vers, m.platform)
 			if err != nil {
 				return err
 			}
 			return pkg
 		}
 
-		pkg, err := d.GetPackage(nil, config.PlatformTuple(), m.Pre)
+		pkg, err := d.GetPackage(nil, m.platform, m.Pre)
 		if err != nil {
 			if !m.Pre && !d.HasNonPrerelease() {
 				for _, cfg := range config.Get() {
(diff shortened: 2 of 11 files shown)


zeroshade:
The failing CI for windows is because windows has different behavior on install, using a registry key rather than a toml file on disk, which is handled in `config/dirs_windows.go`. The `CreateManifest` function in that file is still hardcoding the host platform tuple instead of using the passed in one. Specifically:

```go
// config/dirs_windows.go:321
setKeyMust(dkey, "driver", driver.Driver.Shared.Get(PlatformTuple()))
```

Of course, that won't be sufficient to fix the problem. We don't currently persist the platform tuple in the windows registry with the rest of the manifest. It just has a single `driver` string value in it and assumes that is the default path and will always return the `defaultPath` for any requested tuple. 

So you have to modify the registry usage most likely so that we persist an equivalent to the platformMap into the registry manifest we use.

cmd/dbc/install.go
zeroshade, line 307: you need to also update this method to check the platform in addition to the version so that we don't report a driver is already installed if you're installing for a different platform than what is currently installed.

config/platform.go
```diff
@@ -0,0 +1,69 @@
...
+// Valid platform tuples as in https://dbc-cdn.columnar.tech/index.yaml
+// Please keep up-to-date with the registry index!
+var validPlatformTuples = []string{
+	"linux_amd64",
+	"linux_arm64",
+	"macos_amd64",
+	"macos_arm64",
+	"windows_amd64",
+	"windows_arm64",
+}
```
zeroshade, line 36: At minimum, this should get expanded to cover everything that could be produced by https://github.com/columnar-tech/dbc/blob/d85580c53c8bac9228c62801144f1a533870b7c8/config/config.go#L42-L65

i.e. this is missing `freebsd`, `openbsd`, `linux_x86`, `linux_powerpc64le`, `unknown_wasm64` and others. 

If we don't add those platforms here, then a plain `dbc install` would work on FreeBSD or a x86 host, but `dbc install --platform freebsd_amd64` would get rejected. This list shouldn't be kept in line with the registry index (since we already have handling to report when a driver can't be found for a given platform) instead. We should probably just duplicate the platform validation from the stdlib internal package, something like this:

```go
// validPairs maps GOOS to its allowed GOARCH architectures.
// Sourced directly from Go's official internal/platform list.
var validPairs = map[string]map[string]bool{
	"aix":       {"ppc64": true},
	"android":   {"386": true, "amd64": true, "arm": true, "arm64": true},
	"darwin":    {"amd64": true, "arm64": true},
	"dragonfly": {"amd64": true},
	"freebsd":   {"386": true, "amd64": true, "arm": true, "arm64": true, "riscv64": true},
	"illumos":   {"amd64": true},
	"ios":       {"amd64": true, "arm64": true},
	"js":        {"wasm": true},
	"wasip1":    {"wasm": true},
	"linux": {
		"386": true, "amd64": true, "arm": true, "arm64": true,
		"loong64": true, "mips": true, "mips64": true, "mips64le": true,
		"mipsle": true, "ppc64": true, "ppc64le": true, "riscv64": true, "s390x": true,
	},
	"netbsd":  {"386": true, "amd64": true, "arm": true, "arm64": true},
	"openbsd": {"386": true, "amd64": true, "arm": true, "arm64": true},
	"plan9":   {"386": true, "amd64": true, "arm": true},
	"solaris": {"amd64": true},
	"windows": {"386": true, "amd64": true, "arm": true, "arm64": true},
}

// IsValidPlatform verifies if a GOOS/GOARCH combination is compilation-ready.
func IsValidPlatform(platformTuple string) bool {
    goos, goarch, found := strings.Cut(platformTuple, "_")
    if !found {
        return false
    }
	goos = strings.ToLower(strings.TrimSpace(goos))
	goarch = strings.ToLower(strings.TrimSpace(goarch))

	archs, osExists := validPairs[goos]
	if !osExists {
		return false
	}
	return archs[goarch]
}
```

Obviously doing the swap from darwin -> macos and whatever else we need.

cmd/dbc/install.go
```diff
@@ -470,7 +473,7 @@ func (m progressiveInstallModel) startInstalling(downloaded *os.File) (tea.Model
 	if m.isLocal {
 		driverName := strings.TrimSuffix(
 			strings.TrimSuffix(filepath.Base(m.Driver), ".tar.gz"), ".tgz")
-		parts := strings.Split(driverName, "_"+config.PlatformTuple()+"_")
+		parts := strings.Split(driverName, "_"+m.platform+"_")
```
zeroshade, line 476: this can cause a problem now.

If you did `dbc install ./foo_linux_amd64_1.0.0.tar.gz` on macOS, the manifest would record `foo_linux_amd64` as the whole driver name and record it as a macOS driver unless you explicitly pass `--platform linux_amd64` when doing the local package install. Should we at least attempt to try to derive the platform from the file name first?

cmd/dbc/install_test.go
```diff
@@ -629,3 +629,26 @@ func (suite *SubcommandTestSuite) TestInstallJSON_AlreadyInstalledChecksumFailur
 	suite.Require().NoError(json.Unmarshal(errEnv.Payload, &errPayload))
 	suite.Equal("install_failed", errPayload.Code, "expected install_failed error code")
 }
+
+func (suite *SubcommandTestSuite) TestInstallWithPlatform() {
+	const platform = "linux_amd64"
```
zeroshade, line 634: when this test runs on a linux_amd64 host, it doesn't use the cross-platform branch at all and isn't testing anything useful. Make sure that the tuple we use for this test is guaranteed to differ from `PlatformTuple()` at runtime

client_methods_test.go
```diff
@@ -119,15 +119,25 @@ func TestClientInstall(t *testing.T) {
...
-		manifest, err := c.Install(t.Context(), cfg, "test-driver-1")
+		manifest, err := c.Install(t.Context(), cfg, "test-driver-1", "")
 		require.NoError(t, err)
 		require.NotNil(t, manifest)
 		assert.Equal(t, "test-driver-1", manifest.DriverInfo.ID)
 		assert.NotNil(t, manifest.DriverInfo.Version)
 	})
 
+	t.Run("installs driver for explicit platform", func(t *testing.T) {
+		const platform = "linux_amd64"
```
zeroshade, line 130: same comment as elsewhere, change this so that `platform` is guaranteed to not match the host `PlatformTuple()` at runtime regardless of where this is run.

client_methods_test.go
```diff
@@ -119,15 +119,25 @@ func TestClientInstall(t *testing.T) {
...
 	t.Run("installs driver successfully", func(t *testing.T) {
-		manifest, err := c.Install(t.Context(), cfg, "test-driver-1")
+		manifest, err := c.Install(t.Context(), cfg, "test-driver-1", "")
 		require.NoError(t, err)
 		require.NotNil(t, manifest)
 		assert.Equal(t, "test-driver-1", manifest.DriverInfo.ID)
 		assert.NotNil(t, manifest.DriverInfo.Version)
 	})
 
+	t.Run("installs driver for explicit platform", func(t *testing.T) {
```
zeroshade, line 129: Since the subtests here all share the `tmpDir` and `cfg` this test is running against state left by the preceding "installs driver successfully" test. We should probably give this it's own `t.TempDir()` in this subtest

config/platform_test.go
```diff
@@ -0,0 +1,55 @@
...
+		"darwin_arm64", // wrong OS name
+		"linux", // missing arch
+		"junk", // invalid name
+	    "LINUX_AMD64", // wrong case
+	}
+	for _, tuple := range invalid {
+		var p Platform
+		err := p.UnmarshalText([]byte(tuple))
+		assert.ErrorContains(t, err, "unknown platform")
+		assert.ErrorContains(t, err, "valid values are:")
```
zeroshade, line 48: include `tuple` in the message so that if something fails, we'll at least know which input failed.

Also, add `assert.Empty(t, p)`

config/platform.go
```diff
@@ -0,0 +1,69 @@
...
+
+func ValidPlatformTuples() []string {
+	return slices.Clone(validPlatformTuples)
+}
```
zeroshade, line 40: nothing is calling this, let's drop this. we don't need it.

cmd/dbc/install.go
zeroshade, line 320: we should add the platform to this so that the JSON output (using `--json`) will also get the platform information for what was installed.

zeroshade (changes requested):
Overall this looks good but there's a structural issue with the behavior.

A cross-platform re-install when a version differs would end up orphaning the original installation. You'll have to re-work how we handle re-installation with cross-platform installs.

Currently, dbc assumes only one platform is ever installed at a time for a given driver. So `InstallDriver` below *always* builds a fresh manifest and sets the platform and driverPath, replacing the manifest rather than merging and leveraging the driverMap to hold all the installed platforms. You'll have to change this behavior, and we'll have to decide what we want the behavior to be for the following cases, assuming the user already has Driver D installed, for platform A at version X:

1. User installs Driver D for platform B at version X
2. User installs Driver D for platform A at version Y
3. User installs Driver D for platform B at version Y

Currently, in this PR, all three cases would result in a manifest that only references one platform at one version, potentially leaving behind the extracted tarball from the previous platform without the manifest referencing it for Uninstall to remove. 

@amoeba how do you envision each of the scenarios above *should* result when using the platform option for `dbc install`? 

Conversely, if a user does `dbc uninstall` should it remove for *all* platforms? Or do we need to add a `--platform` option to uninstall as well?

cmd/dbc/install.go
wbeardall: Thanks for the catch on this!

PR #467, opened 2026-08-22
feat: add --all-platforms option for dbc list

wbeardall:
Follow-up to [PR #466](https://github.com/columnar-tech/dbc/pull/466)

## Problem

With the new cross-platform install options proposed in [PR #466](https://github.com/columnar-tech/dbc/pull/466), the following scenario introduces potential user confusion risk:

1. On a fresh machine, with no installed drivers, the user executes `dbc install postgres --platform <not-my-platform>`
2. Later, they want to know if they have a Postgres driver that they can use with ADBC
3. They run `dbc list`. Current `dbc list` does *not* distinguish between the presence of drivers for the host platform vs other platforms, and so prints the following:

```
DRIVER     VERSION LEVEL LOCATION                                            
                                                                             
postgresql 1.12.0  user  /path/to/my/adbc/drivers
```
4. They trust that they can then run ADBC Postgres, and become irate when, at runtime, they *do not have a host-compatible lib*!

## Proposal

We propose to:
1. Add filter-by-default to `dbc list`, so that only drivers which are compatible with the host machine show up in the printed summary
2. Add an opt-in `--all-platforms` flag to `dbc list`, which adds an additional `PLATFORM` column to the printed table, containing the platform tuple for each available driver. Host-compatible drivers are marked with `(*)` (e.g. `linux_amd64 (*)`)

### `defaultPath` configurations

`dbc` `<driver>.toml` config files accept two shared path configurations:

driverMap:
```
[Driver.shared]
linux_amd64 = /path/to/driver
```

defaultPath:

```
[Driver]
shared = /path/to/driver
```
driverMap is trivial; however, for the defaultPath pattern, we assume that, in absence of explicit platform information, the linked driver *is* host-compatible.

### `config/driver_test.go`

- `TestLoadDriverFromManifest` (extended)
  - Asserts `HasPlatform` / `PlatformTuples` for platform-map manifests
- `TestDriverMapDefaultPath` (new)
  - String `shared` → `HasPlatform` true for any platform
  - `UsesDefaultPath` true, `PlatformTuples` empty

### `cmd/dbc/list_test.go`

- `TestListHidesNonHostPlatformDriver`
  - Cross-platform install invisible in default `dbc list`
- `TestListAllPlatformsShowsCrossPlatformDriver`
  - Visible with `PLATFORM` column; `(*)` only when tuple is host
- `TestListAllPlatformsMarksHostDriver`
  - Host install shows `PlatformTuple() + " (*)"` under `--all-platforms`
- `TestListAllPlatformsJSONIncludesPlatform`
  - JSON `platform` field populated when `--all-platforms` is set

### DIFF
--- a/client_methods.go
+++ b/client_methods.go
@@ -247,7 +247,9 @@ func (c *Client) Download(ctx context.Context, pkg PkgInfo) (io.ReadCloser, erro
 }
 
 // Install installs a driver with the given name to the specified configuration.
-func (c *Client) Install(ctx context.Context, cfg config.Config, driverName string) (*config.Manifest, error) {
+// platform selects the driver package platform tuple; when empty, the host
+// platform is used.
+func (c *Client) Install(ctx context.Context, cfg config.Config, driverName string, platform config.Platform) (*config.Manifest, error) {
 	drivers, err := c.Search(ctx, driverName)
 	// Only fail if the driver wasn't found in any registry; partial registry errors
 	// are acceptable as long as we can still locate the target driver.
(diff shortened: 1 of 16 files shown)


zeroshade:
@amoeba what do you think of this solution vs potentially just adding a "platforms" column to the `dbc list` output?

PR #477, opened 2026-08-27
docs: improve appearance of top of index page and add csharp

ianmcook:
- adds C# examples to the control on the index page
  - fixes a tab rendering issue
  - refines the three-step dbc configurator controls with a contained panel, clearer separation, and circled step numbers
  - adds C#/.NET driver manager install commands


amoeba (approved):
LGTM

